PR #56, opened 2021-11-02
Update empty_environment to match buoyant_tethys

chapulina:
* Use same step size
* Remove contact
* Update GUI plugins (which we should eventually move to a GUI config)


mabelzhang (approved):

chapulina:
The failing test on CI should be fixed by #61 

Merging!

PR #57, opened 2021-11-03
Damp pitchmass oscillations.

arjo129:
This PR submits 2 changes. The first is to shift the center of mass of the baselink back and move the center of mass of the vehicle forward. This helps resolve some of the spinning in #47. It also is based on the actual positions of the robot.

The second change made is an increase in distance between the linkand the center of volume. This helps damp oscillations when running `testPitchMass.xml`.

Additionally, this commit updates the pitchmass plots to contain the Error as well as the current Pitch command.

Signed-off-by: Arjo Chakravarty <arjo@openrobotics.org>

Issue #47, opened 2021-10-13
Using the VBS to sink causes the vehicle to translate forwards and go into a spin.


arjo129:
How to reproduce:

In one terminal:
```
ign gazebo -v4 buoyant_tethys.sdf
```

In another:
```
ign topic -t  /model/tethys/buoyancy_engine/ -m ignition.msgs.Double -p "data: 0"
```

## Expected behavior

The vehicle pitches a little bit and then sinks mostly vertically.

## Current behaviour
![Peek 2021-10-13 14-18](https://user-images.githubusercontent.com/542272/137077939-40cd6ed1-3b9b-4f51-93f3-6376201a6c99.gif)
The vehicle sinks but keeps getting pushed backwards, this is likely to some weird numerical issue.

PR #59, opened 2021-11-03
Fix added mass contribution.

arjo129:
I noticed that added mass was acting in the wrong direction. This was after visuallizing the individual forces separately (in this case added mass and damping forces). To be honest it still doesn't fix any of our outstanding issues.


chapulina:
This PR is causing test failures. I'm not sure yet if we should update the tests or if there's something wrong with the PR. One thing I noticed in the tests is that this PR is exacerbating this issue: https://github.com/osrf/lrauv/issues/47

Is there a way we can unit test the plugin so we check what the high-level expectations are? i.e. move 2 vehicles forward, one with added mass enabled and one disabled, and check that the one affected by added mass is slower.

arjo129:
I need to update test expectations. We can extract out the individual components a functions and unit test the functions as is. One of the problems I face is that I can't test the wrenches with the physics engine running (this would be ideal). Otherwise the easiest test is that the direction of the force should always oppose the direction of motion. For some remedy to 47 I would recommend looking at #57.

chapulina:
> I can't test the wrenches with the physics engine running

Can you elaborate? I think it should be possible to test the hydrodynamics isolated in a world without gravity, buoyancy or lift-drag. Apply a force to the vehicle, and see if its resulting velocity is damped as expected?

> the direction of the force should always oppose the direction of motion

+1, I think this should help us detect sign errors

tfoote:
> the direction of the force should always oppose the direction of motion

That's not exactly correct. Only when accelerating. When decelerating it will go in the direction of motion. It should be in the opposite of the direction of the acceleration.

arjo129:
That's correct for the direction of the added mass, but the damping matrix would dominate at higher velocities.

chapulina:
@arjo129 , how would you like to proceed here? Let me know if you need help updating the test expectations.

chapulina (approved):

PR #60, opened 2021-11-03
[QOL improvement] Exit if the docker image does not build

arjo129:
This PR adds a `set -e` command to make sure the script exits if the source fails to build for some reason. The previous behaviour was annoying as it would execute the last good image, this leads to a lot of confusion when debugging.


tfoote (approved):
It's usually more common to set this at the top. But if that already has exit logic lower is fine.

chapulina (approved):

Issue #49, opened 2021-10-15
Investigate weird oscillation in main


arjo129:

braanan:
@chapulina @arjo129 Does this jittery oscillation also happen at depth or is it just on the surface? Is the buoyancy plugin accounting for partially submerged volumes etc? Thanks

chapulina:
> Does this jittery oscillation also happen at depth or is it just on the surface? Is the buoyancy plugin accounting for partially submerged volumes etc?

We have uniform buoyancy everywhere right now, no distinction between water and air until #38 is addressed. So the depth shouldn't make a difference here.

Issue #48, opened 2021-10-14
Document / update coordinates and units


chapulina:
As I'm going through the code, I have various questions about coordinates, units, etc. I'm planning to go over them one by one, documenting them and fixing anything in case it's not correct (hopefully it's all correct though).

- [x] prop omega: https://github.com/osrf/lrauv/pull/76
- [x] rudder angle: https://github.com/osrf/lrauv/commit/86488798e4331118380eb5607e284e19bdf617eb 
- [x] elevator angle: #63 
- [x] mass position: https://github.com/osrf/lrauv/commit/2b07660710f29e71e790e2b2f5fd14f03cddb671
- [x] buoyancy action: #65
- [x] pitch (as defined on missions): https://github.com/osrf/lrauv/commit/2b07660710f29e71e790e2b2f5fd14f03cddb671
- [x] speed (as defined on missions): https://github.com/osrf/lrauv/pull/76

chapulina:
As part of this issue, I started writing tests for the Thruster plugin upstream, because it was originally merged without tests in https://github.com/ignitionrobotics/ign-gazebo/pull/749.

Working on this branch:

https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo5...chapulina/5/thruster_test

chapulina:
Reopening, we still need to sort out the coordinate frames

chapulina:
### What still needs fixing:

- [ ] Model faces West by default (i.e. zero orientation means West), but it should be North to satisfy NED. 
    - [ ] The hardcoded model in `buoyant_tethys` needs to be updated
    - [ ] The spawning mechanism in `empty_environment` needs to be updated
- [ ] Some state reported from simulation to the controller needs updates
    - [ ] `pos_` should be in NED, but it's currently in ESD
    - [ ] `rph_` should be in NED, but it's currently in END
    - [ ]  `posRPH_` should match `rph_` but isn't being populated
    - [ ] `posDot_` should be in NED, but it's currently in ESD
    - [ ] `rateUVW_` should be in FSK, but I'm not sure what it currently is
    - [ ] `ratePQR_` should be in FSK, it isn't currently populated

### What's been done so far:

* https://github.com/osrf/lrauv/pull/81
    * Updates documentation
    * Adds tests
    * Initially attempted to fix frames by adding offsets when spawning the model and then when converting the data back to the controller.
[...]

PR #63, opened 2021-11-04
Elevator example, test and docs

chapulina:
Part of #48 

* Added example executable that controls just the elevator
* Fixed the forward direction in the lift-drag plugin to be -X, because the link's -X is facing the forward direction. But I think that what really matters is the axis, not necessarily the direction.
* Documented the direction of the elevator joint angle

image

* Added a test that checks that the vehicle goes down for a positive elevator angle when propelled forward.

---

I checked manually that the behaviour doesn't seem to be entirely correct though. I'd expect a positive angle of attack to generate lift on the back of the vehicle, so it tilts forward and goes down when propelled forwards. But at low thrust, the vehicle is going up. The higher the thrust, the more it arches downwards. In all cases though, the oscillation of death is present (probably related to #49 ).

### elevator = 0.26 rad, prop omega = 3.0 rad/s

omega3

### elevator = 0.26 rad, prop omega = 10.0 rad/s

omega10

### elevator = 0.26 rad, prop omega = 30.0 rad/s

~~This is the value used in the test~~

omega30

# Update

At @arjo129 's suggestion I used a lower angle of attack on the test, which has less oscillation for higher prop omegas: 

### elevator = 0.15 rad, prop omega = 30.0 rad/s

0 15rad30rads

### DIFF
--- a/lrauv_ignition_plugins/test/test_elevator.cc
+++ b/lrauv_ignition_plugins/test/test_elevator.cc
@@ -40,11 +40,12 @@ TEST_F(LrauvTestFixture, Elevator)
   cmdMsg.set_propomegaaction_(30);
 
   // Rotate elevator counter-clockwise when looking from starboard, which
-  // causes the vehicle to move down
-  cmdMsg.set_elevatorangleaction_(0.8);
+  // causes the vehicle to pitch down
+  cmdMsg.set_elevatorangleaction_(0.27);
 
   // Neutral buoyancy
   cmdMsg.set_buoyancyaction_(0.0005);
+  cmdMsg.set_dropweightstate_(true);
 
   // Run server until the command is processed and the model reaches a certain
   // point vertically
@@ -54,10 +55,21 @@ TEST_F(LrauvTestFixture, Elevator)
     return this->tethysPoses.back().Pos().Z() > targetZ;
   });
 
-  EXPECT_LT(100, this->iterations);
-  EXPECT_LT(100, this->tethysPoses.size());
+  EXPECT_LT(1000, this->iterations);
+  EXPECT_LT(1000, this->tethysPoses.size());
 
   // Vehicle goes down
   EXPECT_GT(targetZ, this->tethysPoses.back().Pos().Z());
+
+  for (int i = 0; i < this->tethysPoses.size(); ++i)
+  {
+    auto pose = this->tethysPoses[i];
+
+    // FIXME(anyone) It goes up a little bit in the beginning
+    if (i < 1000)
+      EXPECT_GT(0.2, pose.Pos().Z()) << i << " -- " << pose.Pos().Z();
+    else
+      EXPECT_GT(0.0, pose.Pos().Z()) << i << " -- " << pose.Pos().Z();
+  }
 }
 
--- a/lrauv_ignition_plugins/test/test_elevator.cc
+++ b/lrauv_ignition_plugins/test/test_elevator.cc
@@ -41,7 +41,8 @@ TEST_F(LrauvTestFixture, Elevator)
 
   // Rotate elevator counter-clockwise when looking from starboard, which
   // causes the vehicle to pitch down
-  cmdMsg.set_elevatorangleaction_(0.27);
+  // Using an angle lower than the stall angle 0.17
+  cmdMsg.set_elevatorangleaction_(0.15);
 
   // Neutral buoyancy
   cmdMsg.set_buoyancyaction_(0.0005);
@@ -55,8 +56,8 @@ TEST_F(LrauvTestFixture, Elevator)
     return this->tethysPoses.back().Pos().Z() > targetZ;
   });
 
-  EXPECT_LT(1000, this->iterations);
-  EXPECT_LT(1000, this->tethysPoses.size());
+  EXPECT_LT(1500, this->iterations);
+  EXPECT_LT(1500, this->tethysPoses.size());
 
   // Vehicle goes down
   EXPECT_GT(targetZ, this->tethysPoses.back().Pos().Z());
@@ -65,8 +66,9 @@ TEST_F(LrauvTestFixture, Elevator)
   {
     auto pose = this->tethysPoses[i];
 
-    // FIXME(anyone) It goes up a little bit in the beginning
-    if (i < 1000)
+    // FIXME(chapulina) It goes up a little bit in the beginning, is this
+    // expected?
+    if (i < 1500)
       EXPECT_GT(0.2, pose.Pos().Z()) << i << " -- " << pose.Pos().Z();
     else
       EXPECT_GT(0.0, pose.Pos().Z()) << i << " -- " << pose.Pos().Z();
--- a/lrauv_ignition_plugins/test/test_elevator.cc
+++ b/lrauv_ignition_plugins/test/test_elevator.cc
@@ -1,0 +1,63 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include <chrono>
+#include <gtest/gtest.h>
+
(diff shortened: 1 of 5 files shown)


lrauv_ignition_plugins/test/test_elevator.cc
```diff
@@ -0,0 +1,75 @@
...
+
+  // Propel vehicle
+  lrauv_ignition_plugins::msgs::LRAUVCommand cmdMsg;
+
+  // Move forward
+  cmdMsg.set_propomegaaction_(30);
+
+  // Rotate elevator counter-clockwise when looking from starboard, which
+  // causes the vehicle to pitch down
+  cmdMsg.set_elevatorangleaction_(0.27);
```
arjo129, line 44: Given the stall angle of the air foil is 0.17 radians as defined here: https://github.com/osrf/lrauv/blob/1181aa3c4515ed131d0994fd4b443c5017144157/lrauv_description/models/tethys_equipped/model.sdf#L118 I don't think we should expect any reasonable behavior if we move the rudder passed that.

arjo129 (changes requested):

lrauv_ignition_plugins/test/test_elevator.cc
braanan: Hi @arjo129 @chapulina, 

Here are notes regarding the LRAUV control surfaces along with a plot of their lift/drag coefficients as a function of angle of attack. As you can see, the control surfaces are still generating lift with minimal drag throughout the nominal operation range of the fins (+/- 15 degrees). Drag increases past the nominal range. As Arjo pointed out, alpha stall is found at the edge of the linear range around +/-0.17 rad (or ~9.7 deg). 

lrauv_lift_drag

**LRAUV control surfaces:**
area: Surface area of the link.

122 cm^2 for one fin.

a0: The initial "alpha" or initial angle of attack. a0 is also the y-intercept of the alpha-lift coefficient curve.

0.0 degrees

cla: The ratio of the coefficient of lift and alpha slope before stall. Slope of the first portion of the alpha-lift coefficient curve.

C_{L\alpha} = 4.13  (alpha in radians)

cda: The ratio of the coefficient of drag and alpha slope before stall.

The coefficient of drag at alpha =0 is 0.03.  The ratio drag/aoa is given below.  Please see Coeffs.png.

C_{D\alpha} = .04/.2 = .2 (alpha in radians)

cp: Center of pressure. The forces due to lift and drag will be applied here.

(3.7, 7.9) cm as shown in the attached diagram OriginalFin.png.  This moves as a function of aoa, as show by the attached diagram cp.png.  Note that cp.png is referenced to mean chord length, not the root chord length.

forward: 3-vector representing the forward direction of motion in the link frame.

[1 0 0]

upward: 3-vector representing the direction of lift or drag.

[0 0 -1]

alpha_stall: Angle of attack at stall point; the peak angle of attack.

9.5 degrees (.17 rad)

cla_stall: The ratio of coefficient of lift and alpha slope after stall. Slope of the second portion of the alpha-lift coefficient curve.

-1.1 (alpha in radians)

cda_stall: The ratio of coefficient of drag and alpha slope after stall.

.030 (alpha in radians)

chapulina: Ohh good point. I see on the controller's config that the rudder angle limit is 15 deg (0.261799 rad), which matches the joint limits. The only stall angle that I found on MBARI's code was the tailcone's stall angle, which is 30 deg. Do you remember where the elevator and rudder stall angles that we're using on lift-drag were taken from? We should document it.

If angles beyond stall result in unreasonable behaviour (at least with our lift-drag plugin), maybe we should limit the joint there?

arjo129: From the above plot. Ben sent it to us on slack.

braanan: I believe the values for the coeffs in the plugin came from the notes above. These notes are from Rob McEwen, our controls engineer. The lrauv-application already limits the angle of the fins to +/- 15 deg, and that's the range we operate with in the field. I guess it's worth noting that the plot above is for the angle of attack and not the fin angle. Still, plenty of lift should be generated past the alpha stall — I think that's also shown in this Gazebo tutorial: http://gazebosim.org/tutorials?tut=aerodynamics&cat=physics

It might be worth revisiting the coeffs values since we there might be a gap between Rob's understanding and what we're doing with these values in the plugin.

chapulina: @braanan posted 2 mins before me :smile: 

It would be nice to double check all these values just to be sure. For example, not sure if by a "fin" we're considering both sides of the elevator or each of them separately. It would also be nice to double check coordinate frames.

> I don't think we should expect any reasonable behavior if we move the rudder/elevator passed that.

@arjo129 , do you expect that because of some limitation on our `LiftDrag` plugin or just in general?

---

As a side note, I had tried some different `cp` values earlier (we currently have `0 0 0` and the simulation crashed. `cp` above is non-zero, so maybe this is one thing for us to look at.

chapulina: Ok, the behaviour looks more reasonable to me at

### elevator = 0.15 rad, prop omega = 30.0 rad/s

0 15rad30rads

But at lower prop omegas, the vehicle still takes off, is that expected?

chapulina: I moved @braanan 's comment to this wiki page: https://github.com/osrf/lrauv/wiki/Lift-drag

Also updated the test to use 0.15 rad.

braanan: Thanks for capturing those notes. Interesting — that looks much better! Guess we'll have to tweak the lift-drag plugin to support the full actuator range — as I mentioned, the vehicles run with the full +/- 15 deg range.

braanan: So I guess the elevators were simply stalling in the previous example?

braanan: 30 rad/s is pretty close to our nominal 1 m/s speed but we typically have good control authority at much lower speeds as well — the vehicle's can easily run at 0.5 m/s and even slightly lower than that

braanan: Can you get the speed through the water?

braanan:
FYI, the source for calculating lift drag for the control surfaces in the MBARI sim can be found here: https://bitbucket.org/mbari/lrauv-application/src/master/Source/simulatorModule/Simulator.cpp#lines-852

lrauv_ignition_plugins/test/test_elevator.cc
chapulina: > Can you get the speed through the water?

* elevator = 0.15 rad, prop omega = 30.0 rad/s
* Plotting XYZ linear velocity in the world frame. It eventually settles around:
    * -0.93 m/s on X
    * -0.01 m/s on Y
    * -0.29 m/s on Z

velocity

braanan: Thanks! Hmm.. are these in the vehicle frame? X velocity magnitude is about right but I'd expect the sign to be positive.

braanan: Ah, oops — I see there in the world frame!

braanan: Nice!

arjo129: So liftdrag plugin, should theoretically be giving the correct force values up to +/- 15 degrees. The liftdrag doesn't support any Cp other than Cp (0 0 0).

chapulina: I captured the gist of this conversation in https://github.com/osrf/lrauv/issues/78

chapulina:
@arjo129 , are there any other changes you'd like to see in this PR? I believe I've addressed all your comments.

arjo129:
So I went on to study why our vehicle is slowly floating up at low velocities:

**At 3 rads^-1, elevator pitch =0.26** 

The vehicle floats up, I was wrong about the hydrodynamics as it is always acting in the correct direction opposing the general direction of motion.

image

Now look at the forces at high velocities:

This is at **30rads^-1, pitch =0.26**
image

For some reason we have a non-zero force from the rudder. Direction of liftdrag and hydrodynamics are both correct. This leads me to believe that the issue is with the hydrostatics and mass distribution of the vehicle.

arjo129 (approved):
I'm OK with approving this PR as I don't think the issues highlighted by @chapulina have much to do with the odd behaviour at low velocities.

chapulina:
> For some reason we have a non-zero force from the rudder...

Interesting, does the vehicle have any yaw then? 

> Note: I am unable to reproduce the oscilatiions here.

You mean the "skipping" from https://github.com/osrf/lrauv/issues/78? Interesting, I have to check if it's my setup then. Does the vehicle stabilize at a downwards pitch for you then?

PR #61, opened 2021-11-04
Increase tolerance of depth and pitch on spawn test

chapulina:
I noticed the spawn test failing on CI. This should make it more robust to transport timing.

PR #64, opened 2021-11-05
change spawn tolerances to be abstracted to a reused variable

tfoote:
From my review comments on #61


chapulina:
See 067cfd98d7e69bd41e053ca19ee06e269c10ea91 and let me know what you think

tfoote:
fixed in upstream pr update

PR #65, opened 2021-11-05
Document buoyancy action / VBS

chapulina:
Part of #48 

* Document `buoyancyAction` and `buoyancyPosition`. Looking at the code history, it looks like at some point cc (cubic centimeter) was used as a unit, but AFAICT, both our plugins and the controller are consistently using cubic meters, so I documented this.
* Made the `LRAUV_example_buoyancy` more flexible so we can test different volumes.
* Checked that `test_buoyancy_action` has correct expectations
* Updated the TODO in the `buoyancy_engine` link, becuase I believe we have a different issue now. 

It would be good to have some eyes on this PR by people who have been involved in the original implementation (@mabelzhang / @arjo129 ), just to make sure I didn't miss anything related to units and the inertial TODO.

### DIFF
--- a/lrauv_description/models/tethys/model.sdf
+++ b/lrauv_description/models/tethys/model.sdf
@@ -273,11 +273,20 @@
       <!-- TODO: Determine location of buoyancy engine -->
       <pose>0.4 0 0 0 0 0</pose>
 
+      <!-- TODO: Remove inertial and collision
+
+         This link was originally given inertia because Gazebo would segfault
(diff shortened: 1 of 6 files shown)


mabelzhang (commented):
Yeah we changed cc to m^3 to be consistent with Ignition units. Initially we used cc because that's how MBARI refers to the volume, both in speech and in code.

I looked over the code, but Arjo should have a look since he did the implementation.

arjo129 (approved):
Thanks! Tested the changes to the example and it works as expected.

PR #67, opened 2021-11-08
[WIP][QOL] Export CSV log files whenever running integration tests.

arjo129:
Often when debugging, if we change one small thing, it leads to a larger impact elsewhere. For instance, I might tweak the values for buoyancy so that say `testPitchMass` has no more oscillations, however that may impact YoYo. Currently, the mission integration tests serve this purpose, but if they fail, it means I have to run the code on my computer again. This PR makes it easier by saving and exporting logs of the application themselves.

Internally it uses https://github.com/actions/upload-artifact to extract and upload the logs.

The log files may be found by clicking on the  _:heavy_check_mark: or :x: > `details` > `summary`_

## Step 1

image

## Step 2

image

## Step 3

image


.github/workflows/integration_tests.yaml
```diff
@@ -13,5 +13,17 @@ jobs:
...
+      run: docker build -t mbari_lrauv_test . -f docker/tests/Dockerfile
+    - name: Create results directory
+      run: mkdir results
+    - name: Run tests
+      run: docker run -v "$(pwd)/results":/results:rw mbari_lrauv_test
+    - name: Get current dir structure
+      run: find .
+    - name: Upload test results.
+      uses: actions/upload-artifact@v2
+      if: always()
```
chapulina, line 26: Is this line needed?

lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
```diff
@@ -171,10 +171,32 @@ class LrauvTestFixture : public ::testing::Test
     return pid;
   }
 
+  public: static void ExportLogs(std::string _target)
+  {
+    int res = system("sudo /home/developer/lrauv_ws/src/lrauv/lrauv_ignition_plugins/plots/unserialize_for_plotting.sh");
```
chapulina, line 176: Is sudo needed?

lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
```diff
@@ -171,10 +171,32 @@ class LrauvTestFixture : public ::testing::Test
     return pid;
   }
 
+  public: static void ExportLogs(std::string _target)
```
chapulina, line 174:
```suggestion
  public: static void ExportLogs(const std::string &_target)
```

lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
```diff
@@ -171,10 +171,32 @@ class LrauvTestFixture : public ::testing::Test
...
+    if (res != 0)
+    {
+      ignerr << "Failed to unserialize plots\n";
+      return;
+    }
+    ignmsg << "Finished unserialize, copying.\n";
+    
+    // WARNING: THIS CAN LEAD TO ARBITRARY CODE EXECUTION
+    auto targetfile = _target + ".csv";
+    auto cmd = std::string("sudo cp -r /home/developer/lrauv_ws/src/lrauv/lrauv_ignition_plugins/plots/missions/tmp/tmp.csv /results/") + targetfile;
```
chapulina, line 187: Is there a way we can configure it not to do this when running the tests locally? These files can quickly start to take up space, and since they're ending up far from the code, they may be hard to track down.

Maybe we can do `export STORE_LOGS=true` on CI?

chapulina (commented):
Ohh I didn't know that was possible in GitHub actions! Sweet!

.github/workflows/integration_tests.yaml
arjo129: Yes. If the above step fails, I still want the logs.

lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
arjo129: 996d8ce

tfoote (commented):
How big are these log files? There's a relatively limited quota of storage for artifacts in actions. We have a total of 500MB for the organization. From the screenshot it looks like it's close to 1MB which means that we only get 500 available before we run out of space. We'll need to have a retention policy so that we don't run out of our quota for the whole organization.

Alternatively can we dump the csv to the console output which is free? And then run the script locally to view after copying and pasting the results from the console log?

arjo129:
I expect them to be ~1mb each. Based on [the docs](https://github.com/actions/upload-artifact/blob/37da8e8d08da8a9370ca3bc9322df4b6c3dc49c3/action.yml#L22) it retains the artifact for 90 days max. It can be tweaked.

lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
arjo129: I'm still iterating on this to figure out the docker permissions.

arjo129:
Our tests have been reworked. Most of the dynamics just work now. This would have been nice to have but we can afford to close this PR.

PR #69, opened 2021-11-10
Clean up README and add SimDaemon

mabelzhang:
As discussed, adding `SimDaemon` to README. Looking at the history, apparently I had it in there before but removed it at some point.

Took the opportunity to clean up the README for a public audience and new project members, since it was originally written for only a few people who already knew what was going on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,28 @@ provided to the public on MBARI's DockerHub (see below), the simulated robot
 can be controlled using the same code executed on the real robot.
 This enables the validation of scientific missions for oceanography research.
 
+## Using Docker for this repository (optional)
+
+Optionally, you may choose to build this repository using Docker, for
+convenience.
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -54,43 +73,35 @@ LRAUV_keyboard_teleop
...
+
+Alternatively, you can use the public Docker image (see below).
 
 The integration assumes that this repository is cloned as a sibling of
 the `lrauv-application` repository, i.e.:
 ```
 <workspace>
 |-- lrauv
-`-- lrauv-application
+|-- lrauv-application
```
chapulina, line 90: Nitpick

```suggestion
└-- lrauv-application
```

chapulina (approved):
Thanks for the cleanup

README.md
chapulina: Tools like `tree` use \` instead of `└` if the latter character is not supported. That's the case for most of my docker containers :cry:

Issue #52, opened 2021-10-27
Ignition Marker improvements for LRAUV


mabelzhang:
Given that all of us (@arjo129 @chapulina and I) have been seeing a big need for Ignition Markers to be more usable for debugging purposes, I'm opening this ticket to keep track of specific features we need in this repo (as opposed to opening separate issues upstream, because there are so many things to make it parallel to RViz Marker. We need to focus on a few for now).

Feel free to add.

- [x] Per-vertex coloring for Marker arrays.
   Why: Performance for visualizing science data. See osrf/lrauv#133. Otherwise we can only visualize a very small number of markers without significant lag, ~400 on my Dell XPS 15 9570.
   Implementation: Refer to [RViz Markers](http://wiki.ros.org/rviz/DisplayTypes/Marker) implementation, which has the colors specified as a vector for vector types including `POINTS`, `LINE_LISTS`, etc. Ignition differs in that the material field is very large, and a vector of that creates a lot of overhead. Need design decision.
- [ ] Arrows
   Why: To debug force vectors etc. I've run into this in our GSoC 2020 project with Martiño too. It's a real limitation.
   Current workaround: One way is to use POINTS and LINE_LISTS in combination. Drawback is you won't have per-vertex coloring - see bullet above.
   Implementation: Refer to [RViz Markers](http://wiki.ros.org/rviz/DisplayTypes/Marker) implementation for parameters to specify sizes for different parts of the arrow.
- [ ] (Optional bonus) Marker textures
  Why: Continuous gradient color visualization with more resemblance to typical marine science visualizations, as opposed to discrete points visualization.
  Implementation: Marker textures have recently been added to RViz2 https://github.com/ros2/rviz/pull/719
  Example usage: https://github.com/gbalke/rviz_shader_tester

chapulina:
I looked a bit into adding arrow markers because I thought it wouldn't take long, and also because I wanted to use axis markers to visualize multiple frames of reference (in the context of #81). It turned out that we may need some deep refactoring in `ign-rendering` to do it, see https://github.com/ignitionrobotics/ign-rendering/issues/493. So I think the easier route may be to publish separate markers with a cylinder and a cone that combined make an arrow :confused:

hidmic:
@arjo129 am I right to say this feature request has been captured upstream?

PR #70, opened 2021-11-10
Science data: lat/lon to cartesian

mabelzhang:
I was going to break https://github.com/osrf/lrauv_private/pull/133 into 3 parts, but turns out I can’t cherry-pick from the original branch in the private repo, because the branch is transferred to the new repo as a single commit.

Just as well, because without the whole thing, you won’t be able to test the visualization anyway. It needs the optimization flags and hacks, otherwise it’ll either lag or be visualized on a scale that the orbiting tools won’t allow you to see.

So it’s just 1 PR, hopefully not too big.

Many performance tricks are added, which take the shape of constant variables. These are needed to see anything at all even just for sanity checks. See Future Work section below.

## To test

```
ign gazebo buoyant_tethys.sdf -v 4
```

Unpause.
Expand “Visualize science data” plugin panel on upper-right.
Click the orange refresh button in the panel.
Try the 3 topics in the drop-down list: `/chlorophyll`, `/salinity`, `/temperature`.
You may need to wait a few seconds for the data to go through the pipeline.
Different colors should show up, looking like this video demo I showed at ROS World:

https://user-images.githubusercontent.com/7608908/141070300-2d769159-089b-474c-8d26-7eee8c446520.mp4

The spam from `TethysCommPlugin.cc` is very obnoxious (~~we need a way to easily disable that, perhaps as a SDF plugin parameter~~ Update: #79), you may want to change the `igndbg` messages in this PR to `ignerr` locally in order to see the debug printouts for visualization.

You should see printouts for the first 10 points for validation:
```
[GUI] [Dbg] [VisualizePointCloud.cc:508] Visualizing 404 markers
[GUI] [Dbg] [VisualizePointCloud.cc:383] First point in cloud (size 92355): -1.22956, -16.1997, -0
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 0 at -5.94342, -14.7471, -0, value 16.3002, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 1 at -5.48038, -13.4784, -0, value 16.4099, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 2 at -6.6232, -13.1163, -0, value 17.3562, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 3 at -7.75394, -12.7543, -0, value 17.108, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 4 at -3.86382, -12.5734, -0, value 17.1428, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 5 at -8.8726, -12.3926, -0, value 17.5717, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 6 at -5.01322, -12.2118, -0, value 16.1063, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 7 at -6.15055, -11.8503, -0, value 17.4024, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 8 at -7.27584, -11.489, -0, value 16.9869, type /temperature, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:489] Added point 9 at -3.39804, -11.3084, -0, value 17.14, type /temperature, dimX 0.38088
```

## Future Work

There are still a number of necessary things to do.
Some are currently worked around. Some are hacked around.

World origin lat/long:

- Visualization needs to poll from `WorldCommPlugin` the first spawned vehicle position, in case the user uses `empty_environment.sdf` instead of `buoyant_tethys.sdf`. The former allows a vehicle to be spawned at any lat/long. The latter has the lat/long hardcoded. Currently, the visualization only works with the latter and will have the wrong Cartesian coordinates for the former. It might be done with a service, but WHEN is the question. It cannot be done in `Configure()`, because the user might wait 10 minutes to spawn a vehicle. It might need to be done every once in a while.
  (Update: This might be done by #77 and 3f6edfc, but it needs to be tested with `empty_environment.sdf` and a new vehicle being spawned.)
- After getting that new lat/long, all the science data positions, which would have already been read and converted to Cartesian, need to be transformed to the new lat/long that the world origin is associated with.
- [x] The static world lat/long is hardcoded right now. This should be read from the SDF. (Update: #77 and 3f6edfc)

Sensor coordinates:
- [x] Sensor position needs to be converted from spherical coordinates to Cartesian, in order to be associated with the science data positions, which are now converted to Cartesian in this PR. This PR is only for science data visualization, sensors are out of scope and need to be in a future PR. (Update: #83)

Visualization:
- Per-vertex coloring for Markers as tracked in #52. This should give better performance. Performance is the main bottleneck that requires all the flags tagged `Performance trick` in the code.

- Orbiting tool improvements for larger worlds. Currently, the markers are scaled down to abysmally small scales, in order to get them into view, just for our debugging purposes. Otherwise there is no way to even see where the data are for sanity checks.


lrauv_ignition_plugins/src/WorldCommPlugin.cc
```diff
@@ -104,15 +105,22 @@ void WorldCommPlugin::Configure(
   this->createService = "/world/" + topicWorldName + "/create";
   this->setSphericalCoordsService = "/world/" + topicWorldName
     + "/set_spherical_coordinates";
+
+  // Advertise service for world origin in spherical coordinates
+  this->getWorldOriginSphericalService = "/world_origin_spherical";
+  this->node.Advertise(this->getWorldOriginSphericalService,
+    &WorldCommPlugin::WorldOriginSphericalService, this);
```
chapulina, line 112: See https://github.com/osrf/lrauv/pull/77 for a way to detect that the spherical coordinates changed without a service.

mabelzhang: Thank you! Merged so that I can update this PR to it

mabelzhang: Incorporated into this PR 3f6edfc

mabelzhang:
test_spawn was hanging in the checks.
I disabled science sensors in `empty_environment.sdf`, since we don't shift all the data to the dynamic new world lat/long yet. So it's not meaningful to load science sensors in that world.

Not sure why it was hanging though. It hangs locally for me too. I didn't look closely at what that test is checking.

lrauv_ignition_plugins/src/WorldCommPlugin.cc
chapulina: I saw you resolved this comment; are you planning on removing this service? I think it isn't needed, right?

mabelzhang: Oh I forgot to remove it in WorldCommPlugin! Oops thanks for keeping track. Removed in 58af8e8
PR ready for another pass. I rebased my interpolation branch on this PR, so it'd be nice to get this one in.

chapulina:
I have some suggestions on #85 so we avoid hardcoding specific science data

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -417,6 +515,72 @@ void ScienceSensorsSystemPrivate::ReadData()
...
+          << latLon.value().LatitudeReference().Degree() << ", "
+          << latLon.value().LongitudeReference().Degree() << std::endl;
+
+      this->worldOriginSphericalCoords = latLon.value();
+
+      // Extract world origin in (lat, long, elevation) from spherical coords
+      this->worldOriginSphericalPos = ignition::math::Vector3d(
+        this->worldOriginSphericalCoords.LatitudeReference().Degree(),
+        this->worldOriginSphericalCoords.LongitudeReference().Degree(),
+        0);
```
chapulina, line 540: Any reason to hardcode `0` here?

```suggestion
        this->worldOriginSphericalCoords.ElevationReference());
```

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -417,6 +515,72 @@ void ScienceSensorsSystemPrivate::ReadData()
...
+      // Convert spherical coordinates to Cartesian
+      this->sphCoord = ignition::math::SphericalCoordinates(
+        this->worldOriginSphericalCoords.Surface());
```
chapulina, line 543: It's not clear to me what's the intended difference between `sphCoord` and `worldOriginSphericalCoords`. What this is doing is:

1. Convert `worldOriginSphericalCoords` to `worldOriginSphericalPos`
2. Convert `worldOriginSphericalPos` to `sphCoord`

Are they supposed to be different?

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -99,13 +118,36 @@ class tethys::ScienceSensorsSystemPrivate
...
+
+  /// \brief Mutex for writing to world origin association to lat/long
+  public: std::mutex mtx;
+
+  /// \brief Spherical coordinates of world origin. Can change at any time.
+  public: ignition::math::SphericalCoordinates worldOriginSphericalCoords;
+
+  /// \brief World origin in spherical position (latitude, longitude,
+  /// elevation), angles in degrees
+  public: ignition::math::Vector3d worldOriginSphericalPos = {0, 0, 0};
```
chapulina, line 137: Do we need to store this variable? I don't see it being used anywhere besides right after it is set.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -417,6 +515,72 @@ void ScienceSensorsSystemPrivate::ReadData()
...
+      // Convert spherical coordinates to Cartesian
+      this->sphCoord = ignition::math::SphericalCoordinates(
+        this->worldOriginSphericalCoords.Surface());
+      this->worldOriginCartesianCoords =
+        this->sphCoord.LocalFromSphericalPosition(
+          this->worldOriginSphericalPos);
+
+      this->worldSphericalCoordsInitialized = true;
+
+      //TODO(chapulina) Shift science data to new coordinates
```
chapulina, line 550: I think this TODO can be moved into `ShiftDataToNewSphericalOrigin`

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -573,16 +748,38 @@ void ScienceSensorsSystem::PostUpdate(const ignition::gazebo::UpdateInfo &_info,
...
+
     // For each sensor, get its pose, search in the octree for the closest
     // neighbors, and interpolate to get approximate data at this sensor pose.
     for (auto &[entity, sensor] : this->entitySensorMap)
     {
       // Sensor pose in lat/lon, used to search for data by spatial coordinates
+      // TODO convert to Cartesian
       auto sensorLatLon = ignition::gazebo::sphericalCoordinates(entity, _ecm);
+      /*
+      // TODO DEBUG what is the sensor attached to?? World? Not robot?
```
chapulina, line 778: Are you getting weird values here? Can these debug statements be removed?

chapulina (commented):
Working for me, I just have some minor comments / questions / suggestions

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -515,11 +682,25 @@ void ScienceSensorsSystem::Configure(
   this->dataPtr->cloudPub = this->dataPtr->node.Advertise<
       ignition::msgs::PointCloudPacked>("/science_data");
 
-  this->dataPtr->node.Advertise("/science_data",
+  this->dataPtr->node.Advertise("/science_data_srv",
```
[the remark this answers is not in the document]
mabelzhang: I think I just need to change it back. I used a different name because while trying to understand the code the first time some weeks ago, I was confused because they're the same name, but changing a string didn't change the one I wanted. Maybe it's better to declare a variable and comment that it's used for both the topic and the service.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -751,10 +954,17 @@ ignition::msgs::PointCloudPacked ScienceSensorsSystemPrivate::PointCloudMsg()
   msg.set_row_step(pclPC2.row_step);
   msg.set_is_dense(pclPC2.is_dense);
 
-  // FIXME: Include more than position data
   msg.mutable_data()->resize(pclPC2.data.size());
   memcpy(msg.mutable_data()->data(), pclPC2.data.data(), pclPC2.data.size());
 
+  // Populate float arrays for actual science data
+  *this->tempMsg.mutable_data() = {temperatureArr[this->timeIdx].begin(),
+    temperatureArr[this->timeIdx].end()};
```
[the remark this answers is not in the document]
mabelzhang: Can we do that in #88? I'd like to get this in since #83 depends on this.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -417,6 +515,72 @@ void ScienceSensorsSystemPrivate::ReadData()
...
 }
 
+/////////////////////////////////////////////////
+void ScienceSensorsSystemPrivate::UpdateWorldSphericalOrigin(
+  ignition::gazebo::EntityComponentManager &_ecm)
+{
+  // Lock for changes to worldOriginSpherical* until update complete
+  std::lock_guard<std::mutex> lock(mtx);
+
+  if (!this->worldSphericalCoordsInitialized)
```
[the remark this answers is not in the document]
mabelzhang: Actually, we cannot return right away. That's actually a logic error due to incorrect braces. Thanks for catching that.

The if-statement is there because `UpdateWorldSphericalOrigin()` is called from two places, one in `Configure()` (because otherwise no data is shown), and the other in `PreUpdate()`. In the former case, the variable `worldSphericalCoordsInitialized` is false, and in the latter case, it is true. BUT, in both cases, `ShiftDataToNewSphericalOrigin()` needs to be called. So the `ShiftDataToNewSphericalOrigin()` call is supposed to go outside the if-statement.

That hadn't been caught because `ShiftDataToNewSphericalOrigin()` is not coded yet! So both cases the behavior is the same.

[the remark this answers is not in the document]
mabelzhang: I pulled the function call out of the if-statement (to be committed, will post hash), but the logic might need to be changed once the function is coded. The logic is a combination of `this->worldSphericalCoordsInitialized` and `this->initialized`, which needs to be thought about and tested after `ShiftDataToNewSphericalOrigin()` is implemented.

Really, I just put the calls in there for whoever that implements `ShiftDataToNewSphericalOrigin()` to be aware of these different circumstances under which `UpdateWorldSphericalOrigin()` may be called. The call to `UpdateWorldSphericalOrigin()` in `PreUpdate()` needs to be tested at the same time too, because that condition has also never happened so far. It only happens if a vehicle is spawned.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: Because of this https://github.com/osrf/lrauv/pull/70#discussion_r755672172 , I will leave this here, because this `if` logic may need to be changed.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: This block is removed in #83, where the sensor position is converted to Cartesian and tested. In this PR, the sensor position couldn't be tested until we have #83. I'll leave it because otherwise this PR would be going forward in time.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: I print this out in `ReadData()` in #83. That might be another forward-merge thing. I think this is a useful variable that I'll be using for debugging until all the sensor stuff is finalized. We're still pretty far from that, at least 2 more PRs, before actually testing the missions.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: There's something weird about the conversion. I don't know if it's because I'm using the function wrong, or we're not using Z the way the function assumes.

The depth conversion result looks wrong if I give it the actual elevation, in that it doesn't give us `-depth` as z. Since we're treating z = 0 as surface, and z is supposed to be `-depth`, I just do the lat/long transformation with elevation 0, and then put in the z manually. That is done in `ConvertLatLonToCart()` (I use it more than once in newer code in #83) as well as here.

Short of a fix, I can add a TODO linking to this comment.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: They gave me different results. I don't know why, but I had to create a new one for the markers to be at reasonable places. The difference is that `worldOriginSphericalCoords` contains everything about the world origin, and `sphCoord` is a blank object that only contains `Surface()`, which is probably wrong, but it gives results at the origin, which there should be, since the origin is set to be at a lat/long copied straight from the CSV.

I tried it again with this fix in both places in the code, mentioned here https://github.com/osrf/lrauv/pull/70#discussion_r755679895 , and the former is still giving positions like this:
```
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 0 at -440.367, 365.239, -0, value 0.922332, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 1 at -439.904, 366.508, -0, value 0.437973, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 2 at -441.047, 366.87, -0, value 0.104822, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 3 at -442.178, 367.232, -0, value 0.0355906, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 4 at -438.288, 367.413, -0, value 0.268712, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 5 at -443.296, 367.594, -0, value 0.426041, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 6 at -439.437, 367.775, -0, value 0.527025, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 7 at -440.574, 368.136, -0, value 0.207158, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 8 at -441.7, 368.497, -0, value 0.954847, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 9 at -437.822, 368.678, -0, value 0.0730943, dimX 0.38088
```

`sphCoords` gives points near the origin, which I set in the SDF to an exact lat/long that has a data point
```
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 0 at 0.989183, -3.98017, -0, value 0.922332, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 1 at 1.45222, -2.71143, -0, value 0.437973, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 2 at 0.309398, -2.34932, -0, value 0.104822, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 3 at -0.821341, -1.98739, -0, value 0.0355906, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 4 at 3.06878, -1.80649, -0, value 0.268712, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 5 at -1.94, -1.62567, -0, value 0.426041, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 6 at 1.91938, -1.44486, -0, value 0.527025, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 7 at 0.782051, -1.08337, -0, value 0.207158, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 8 at -0.343238, -0.722056, -0, value 0.954847, dimX 0.38088
[GUI] [Dbg] [VisualizePointCloud.cc:477] Added point 9 at 3.53456, -0.541467, -0, value 0.0730943, dimX 0.38088
```

I will say at the time this was coded, I just needed something, anything, to show up at the origin, so a quick hack was needed. So while `sphCoord` looked wrong to me, I didn't have time to fix it.

Here's what I would propose. Right now, since #83 is pending on this, I would propose to delay the fix again, so that #83 isn't blocked. I know that's not good, but I also suspect after #83 and another PR to do the interpolation, then we can actually test the sensor locations and maybe figure out why this is wrong.

Very honestly, right now, there is still realistically nothing to test this code. We see some colors, but we really have no idea whether they're correct wrt neither lat/ong nor Cartesian! So all of these PRs are just code that won't be tested until a few weeks later.

mabelzhang: TODO added in c287169

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: TODO added in c287169

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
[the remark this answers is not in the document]
mabelzhang: TODOs added in c287169

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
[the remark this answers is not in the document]
mabelzhang: Variable added in c287169

mabelzhang:
Ready for another pass.

I know I'm pushing to delay some fixes, which is not good style, but I'm really anxious to get this in, since #83 depends on this, and I'm working on another one depending on that. I really don't want to chain up more than 3 PRs. We can't run the acceptance test until all 3 PRs mentioned are in, and we only have 3 weeks until we said we wanted to finish the milestone.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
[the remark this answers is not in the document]
chapulina: Absolutely, I just wanted to make sure I mentioned it.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
chapulina: How can I reproduce the failure? I see `buoyant_tethys.sdf` is currently setting the `<elevation>` to zero, so this change here shouldn't make a difference.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
chapulina: > I would propose to delay the fix again, so that #83 isn't blocked

I'm ok fixing this incrementally, my only concern is that the positions could be completely wrong right now. Would that affect how #83 is implemented?

> right now, there is still realistically nothing to test this code

I think we can come up with some tests, we just need to create some dummy data that we know exactly where it should go. In fact, I think we should be working against this kind of data before we try to load the real datasets, because it's much easier to create different scenarios to test.

My suggestion would be to add more datapoints to [dummy.csv](https://github.com/osrf/lrauv/blob/main/lrauv_ignition_plugins/data/dummy.csv) and check that they behave as expected. I'll do that now locally here to help me wrap my head around this PR.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
chapulina: I think it's just holding duplicate data that's already in `worldOriginSphericalCoords`, right? Couldn't you just print the values from that? Keeping the same information in 2 places is prone to error because we need to keep them in sync.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -99,13 +118,36 @@ class tethys::ScienceSensorsSystemPrivate
...
+  /// \brief Spherical coordinates of world origin. Can change at any time.
+  public: ignition::math::SphericalCoordinates worldOriginSphericalCoords;
+
+  /// \brief World origin in spherical position (latitude, longitude,
+  /// elevation), angles in degrees
+  public: ignition::math::Vector3d worldOriginSphericalPos = {0, 0, 0};
+
+  /// \brief World origin in Cartesian coordinates converted from spherical
+  /// coordinates
+  public: ignition::math::Vector3d worldOriginCartesianCoords = {0, 0, 0};
```
chapulina, line 141: I don't understand this variable. Isn't the world origin in Cartesian always zero? What does a non-zero cartesian origin mean?

chapulina:
> I was going to break osrf/lrauv_private#133 into 3 parts, but turns out I can’t cherry-pick from the original branch

Sorry, that's on me.

I do think it would be beneficial to break this PR into smaller ones though, because there's a lot going on here. It will help with review. I can help divvy-up the work if you want.

mabelzhang:
Per discussion offline, #93 breaks off the visualization part of this PR, most of which we already looked at in #85.
After that's merged, this PR should only contain the lat/long to Cartesian conversion.

tfoote:
In testing for #83 I was unable to to collapse the widget after I expanded it to the side bar. I don't know if that's related to this or generally an issue

sidebar:
Screenshot from 2021-11-29 12-54-40


clicked the ^ button to collapse it the display goes away but not the side bar. 

Screenshot from 2021-11-29 12-54-47




And when I expanded it to full screen it overlapped the 3D visualization tab. 

Screenshot from 2021-11-29 12-53-13

chapulina:
> clicked the ^ button to collapse it the display goes away but not the side bar.

That's expected. The side bar takes the full vertical height, even if it doesn't have enough plugins to fill it.

> And when I expanded it to full screen it overlapped the 3D visualization tab.

The plugins can either be floating on top of each other, or docked into the side bar. When it's floating like this, you can just drag it somewhere else.

We don't have support for more complex docking arrangements on Ignition GUI, mainly because that was not available out-of-the-box for QtQuick when we started.

chapulina:
@mabelzhang , thanks for breaking the PR apart.

I have an alternative implementation for the coordinate conversion on #98. Let me know what you think.

mabelzhang:
Closing in favor of #98
Summary: This PR is replaced by #93 and #98

PR #73, opened 2021-11-11
Terminal velocity test

arjo129:
This test evaluates whether the hydrodynamic plugin successfully performs damping when a thrust is applied. Based on previous discussion with MBARI at 300rpm, we should be moving at 1ms^-1. This test checks this behaviour. Furthermore, by starting 4 vehicles out in different positions and orientations, this test makes sure that the transforms of the hydrodynamics plugin are correct.


lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,268 @@
...
+{
+  using namespace ignition;
+
+  transport::Node node;
+  auto pub =
+      node.Advertise<msgs::Double>(
+          "/model/" + _name + "/joint/propeller_joint/cmd_pos");
+
+  msgs::Double thrustCmd;
+  thrustCmd.set_data(-6.7);
```
chapulina, line 58: Can you add a comment explaining where this value comes from? The text above mentions 300 rpm, and I'm not sure how we get from that to -6.7 Newtons.

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,268 @@
...
+
+  for (int i = 0; i < 3; i++)
+  {
+    pub.Publish(thrustCmd);
```
chapulina, line 62: Are you publishing many times because some are lost? I found that sleeping a bit while checking `pub.HasConnections()` helps with that.

chapulina (commented):
I think this is a good integration test. I like it how it tests different directions.

With so many plugins involved, it's hard to tell how much each of them is contributing to the deacceleration though.

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: 8413beb

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: 8413beb

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,278 @@
...
+
+  transport::Node node;
+  auto pub =
+      node.Advertise<msgs::Double>(
+          "/model/" + _name + "/joint/propeller_joint/cmd_pos");
+
+  msgs::Double thrustCmd;
+
+  // 300 rpm -> 31.42rads^-1
+  //   -> (31.42rads^-1)^2 * 0.004422 (thrust coeff) * 1000 (fluid density)
```
chapulina, line 60: I'm proposing we change the coefficient in #76 because the current value only works due to an error on `TethysComm`. See more info on that PR's description.

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,278 @@
...
+
+//////////////////////////////////////////////////
+void commandVehicleForward(const std::string &_name)
+{
+  using namespace ignition;
+
+  transport::Node node;
+  auto pub =
+      node.Advertise<msgs::Double>(
+          "/model/" + _name + "/joint/propeller_joint/cmd_pos");
```
chapulina, line 55: I recommend using 

```suggestion
          "/model/" + _name + "/joint/propeller_joint/cmd_thrust");
```

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,278 @@
...
+
+  fixture->OnPostUpdate(
+      [&](const gazebo::UpdateInfo &_info,
+          const gazebo::EntityComponentManager &_ecm)
+      {
+        auto worldEntity = gazebo::worldEntity(_ecm);
+        gazebo::World world(worldEntity);
+
+        vehicle1 = world.ModelByName(_ecm, "tethys");
+        if (gazebo::kNullEntity != vehicle1)
```
chapulina, line 147: Should these be `EXPECT_NE`s?

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,278 @@
...
+        }
+      });
+  fixture->Finalize();
+
+  commandVehicleForward("tethys");
+  commandVehicleForward("tethys2");
+  commandVehicleForward("tethys3");
+  commandVehicleForward("tethys4");
+
+  // Check that vehicles don't exist
```
chapulina, line 236: I'm confused :thinking:

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,278 @@
...
+        gazebo::World world(worldEntity);
+
+        vehicle1 = world.ModelByName(_ecm, "tethys");
+        if (gazebo::kNullEntity != vehicle1)
+        {
+          auto baselink = gazebo::Model(vehicle1).LinkByName(_ecm, "base_link");
+          gazebo::Link link(baselink);
+          auto velocity = link.WorldLinearVelocity(_ecm);
+          auto pose = link.WorldPose(_ecm);
+          if(!pose.has_value() || !velocity.has_value())
```
chapulina, line 153: Same here, I think these should be `ASSERT_TRUE`s

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,278 @@
...
+  auto pub =
+      node.Advertise<msgs::Double>(
+          "/model/" + _name + "/joint/propeller_joint/cmd_pos");
+
+  msgs::Double thrustCmd;
+
+  // 300 rpm -> 31.42rads^-1
+  //   -> (31.42rads^-1)^2 * 0.004422 (thrust coeff) * 1000 (fluid density) 
+  //      * 0.2m ^ 4 (prop diameter) = 6.9857
+  thrustCmd.set_data(-6.9857);
```
chapulina, line 62: Note that the sign will change after #76

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: See #90

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: 4fa9464

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: 4fa9464

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: 4fa9464

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: 3202298

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
arjo129: 3202298

lrauv_ignition_plugins/test/test_hydrodynamics_equilibrium_velocity.cc
```diff
@@ -0,0 +1,253 @@
...
+  EXPECT_NEAR(
+    velocitiesV1.rbegin()->Z(), 0, 1e-3);
+
+  // Acceleration should always be decreasing.
+  checkDamping(velocitiesV1);
+  checkDamping(velocitiesV2);
+  checkDamping(velocitiesV3);
+  checkDamping(velocitiesV4);
+
+  // Rotations should not have changed much throuh the course of the test
```
chapulina, line 248:
```suggestion
  // Rotations should not have changed much through the course of the test
```

chapulina (approved):
Nice! I think some expectations may end up being flaky, but we can deal with it if / when it happens.

Issue #75, opened 2021-11-11
Update Thruster to also accept angular velocity commands


chapulina:
The thruster plugin, both the one in this repo and the upstreamed one, take a force in Newtons as input, and uses that to calculate the theoretical angular velocity of the propeller with this formula:

`omega = sqrt(thrust / (fluid_density * thrust_coefficient * propeller_diameter ^ 4))`

Since MBARI's controller sends commands in rad / s, we should consider updating the thruster plugin to alternatively take angular velocity commands. Right now, this is what we do:

1. Controller sends `propOmega` in `rad/s`
2. `TethysComm` converts `rad/s` to `N` using a formula adapted from Fossen's book (with values hardcoded in the plugin)
3. `Thruster` receives the `N`, applies that force to the vehicle, and uses the inverse formula to compute the desired `rad/s` and spin the propeller (with values configurable from SDF)

If `Thruster` accepted `rad/s`, we would only need to do the conversion in one place, the `Thuster` plugin, which already offers a way to configure parameters (`fluid_density`, `thrust_coefficient`...) from SDF.

PR #74, opened 2021-11-11
Use upstream's Thruster plugin

chapulina:
Part of #37.

~~I tested this locally with `ign-gazebo`'s branch `chapulina/6/thruster_cherry`, which is https://github.com/ignitionrobotics/ign-gazebo/pull/1190 cherry-picked to Fortress.~~

The upstream PR has been merged and ported to `ign-gazebo6`, so this PR should work with `ign-gazebo6` from source. A release is coming up soon too.

All tests passed for me locally. ~~Let's wait until that PR is merged and released to merge this one.~~ They also pass on CI. So I think we don't need to wait for that PR to merge this unless we really want to make sure we use velocity control.

PR #76, opened 2021-11-12
Document propOmega, flip thruster, fix coefficient

### LINKED ISSUE #20: [Thruster plugin] Flip sign direction.
Flip the sign direction of the thruster plugin/XML of joint axis to make it move forward when we give a positive thrust command.

chapulina:
* Builds on top of #74 
* Requires https://github.com/ignitionrobotics/ign-gazebo/pull/1190
* Last part of #48
* Closes #20 

---

The propeller joint's axis was pointing to the vehicle's back, which caused the vehicle to be propelled forward for negative thrust forces. This PR flip the axis so that clockwise rotations when looking from the back cause the vehicle to move forward:

image

thruster

* Documented `propOmega`
* Flipped the propeller joint, and flipped the commands accordingly
* Fixed the conversion from angular velocity to force. The formula on `TethysComm` was using 0.0016 m for the propeller diameter, but the `Thruster` uses 0.2 m, taken from the SDF. I updated `TethysComm` to use 0.2 m, and that made the vehicle extremely fast. So I reduced the coefficient 100x to compensate.
    * As described on #75, we're converting from ang vel to force and back. It would reduce errors if we only converted once.
* Updated a few test expectations to the new thrust. In particular, yoyo is spinning a bit faster now. I also took the opportunity to add more checks to yoyo, ~~and make the tolerances tighter (from 12 m  / 1.57 rad to 1.5 m / 0.1 rad!!!)~~ and remove the specific trajectory check, because it was too hard to maintain and failures weren't useful :yo_yo:


arjo129 (approved):
Thanks for these changes. The PR Looks Good To Me.

PR #77, opened 2021-11-12
Science sensors: detect when spherical coordinates are set

chapulina:
@mabelzhang , here's the approach I mentioned that we can use to detect when the spherical coordinates are set.

If you run `buoyant_tethys.sdf`, you'll see the warning message at startup. If you run `empty_environment.sdf`, you'll see it when the first vehicle is spawned.

Let me know what you think!


mabelzhang (approved):
Thanks! Tried it, will merge and update in my PR.

PR #79, opened 2021-11-13
Toggle TethysCommPlugin debug printout

mabelzhang:
Because I can't see other printouts amidst this spam (and it was my spam too).

### DIFF
--- a/lrauv_description/models/tethys_equipped/model.sdf
+++ b/lrauv_description/models/tethys_equipped/model.sdf
@@ -151,6 +151,7 @@
         <namespace>tethys</namespace>
         <command_topic>tethys/command_topic</command_topic>
         <state_topic>tethys/state_topic</state_topic>
+        <debug_printout>0</debug_printout>
       </plugin>
       <plugin
         filename="HydrodynamicsPlugin"
--- a/lrauv_ignition_plugins/src/TethysCommPlugin.cc
+++ b/lrauv_ignition_plugins/src/TethysCommPlugin.cc
@@ -605,7 +605,7 @@ void TethysCommPlugin::PostUpdate(
   this->statePub.Publish(stateMsg);
 
   if (this->debugPrintout &&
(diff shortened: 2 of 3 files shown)


chapulina (approved):
I like the idea, but I'd personally prefer something that isn't hardcoded and can be more easily toggled from the command line, such as an environment variable. Something like this:

```
      std::string debugEnv;
      ignition::common::env("TETHYS_COMM_DEBUG", debugEnv);
      this->debugPrintout = (debugEnv == "true");
```

Then run:

```
TETHYS_COMM_DEBUG=false ign gazebo buoyant_tethys.sdf
```

mabelzhang:
I see the reasoning behind not hardcoding. The SDF parameter is nice practically because it's 1 bit to change, 0 or 1.

Environment variables are more words to remember. I'd likely default it to false, and go look for the variable name in the README to set it to true.

The other nice thing about doing it via SDF is that it checks true/false and 0/1, so the code doesn't need to check for cases and numeric vs. string.

How often are you guys using the printout? I don't look at them anymore. They're useful for debugging low-level controls, communications, and time syncing. I haven't been doing those things.

chapulina:
I can see the advantages of SDF. It's just a bit more cumbersome to always keep in mind that the SDF file has a change I don't want to commit... While an env var is contained in my own workspace and doesn't get leaked through the source code.

> How often are you guys using the printout?

I mostly run with `-v 3` so I don't see those anyway :sweat_smile: I guess they can be useful when debugging test failures on CI, though I never used them.

mabelzhang:
Yeah, I know what you mean about not committing things - that was the point I was thinking of too.

But okay, then I think no one is really using those printouts at the moment. I might use them soon when I start to debug the missions again, to check the science sensor data, but for now, it sounds like no one will need to change that SDF parameter. So I'll leave it. We can revisit the env var later if it gets annoying for people (likely me again).

Thanks!

PR #82, opened 2021-11-17
Strip and add back components

arjo129:
This is a "live blog" of myself trying to strip the vehicle down and add components in to determine the cause of the oscillations in the vehicle. The rationale for doing this is to make sure that we can isolate the systems responsible for Hydrostatic instability. I will start with a rigid body with an off-center mass. 

Right now only the buoyancy plugin is enabled.

Requires https://github.com/ignitionrobotics/ign-gazebo/issues/1211

The vehicle in cd7990ae4c8fdd21b9774fb84f8318dbf57b1101 is stable with no oscillation.

### DIFF
--- a/lrauv_description/models/tethys/model.sdf
+++ b/lrauv_description/models/tethys/model.sdf
@@ -9,7 +9,7 @@
     <link name="base_link">
       <inertial>
         <!-- to offset battery CoM before getting real parameters for vehicle without battery -->
-        <pose>-0.1181 0 0 0 0 0</pose>
+        <pose>-0.119779049 0 0 0 0 0</pose>
         <!-- 146.5671 subtracted by battery mass -->
         <!--<mass>114.8364</mass>-->
         <mass>146.5671</mass>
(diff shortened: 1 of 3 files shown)


arjo129:
Adding the drop weight in 14f8267 leads to oscillations. :(

chapulina:
> Adding the drop weight in 14f8267 leads to oscillations. :(

The weight is centered on the model, so its CoM isn't aligned with `base_link`'s CoM. Its inertia also has some off-diagonal terms. So maybe this is expected?

chapulina:
> The weight is centered on the model, so its CoM isn't aligned with base_link's CoM. 

Here's the offset on the CoMs:

com_drop

The model sinks for me on 14f8267 though, it doesn't oscillate

arjo129:
That is correct. Are you using the branch in ignitionrobotics/ign-gazebo#1211 which "solves" the sinking.

arjo129:
Will need to tweak the center of mass of the main vehicle. Also our visual needs to be offset.

arjo129:
So I moved the center of mass of the main body forward but kept the same center of buoyancy in 0f271f2. The craft is once again stable.

arjo129:
Closing this PR in favor of #89

PR #83, opened 2021-11-17
Nearest neighbor search for 8 points on 2 z planes as inputs for interpolation

mabelzhang:
~~Depends on #70.~~ PR on trilinear interpolation to follow.
Anyone free can review. This PR is pretty independent, mainly matrix algorithm.

Two substantial functions are added. To keep PRs easier to review, I thought this should go in before the actual trilinear interpolation in a separate PR.

What this PR does, from major to minor:
- Two new functions, `FindInterpolators()` and `SearchInDepthSlice()`, compute the inputs to go into trilinear interpolation (see algorithm below)
- Convert sensor location from lat/long to Cartesian (this is what depends on #70, for the conversion function, and for the converted sensor position to make sense with the science data, which were already converted)
- Fix a seg fault I previously had in this branch in the private repo
- Decrease the number of times sensor location is updated and interpolation is called. These clog up the pipeline I think, causing science visualization to never show up. Interpolation doesn't need to run for every sensor in every PostUpdate(). It really only needs to happen when the robot has moved far enough to warrant a sensor location update, and only once for the entire vehicle (as opposed to every sensor).

## Algorithm for 8-neighbors search for interpolation

The inputs to trilinear interpolation are 8 conceptual "nearest neighbors," but not precisely NNs, because they are required to be on 2 distinct z slices, in order to be useful for trilinear interpolation.

To find the 8 NNs, I cannot just look for 2 NNs directly, because they just ended up on the same z slice and were useless for interpolation.

This is the current algorithm. It makes liberal use of the raw `Eigen::MatrixXf` underlying `pcl::PointCloud`:
1. Find 1st NN.
2. Create a z slice of all points in cloud matching the z of the 1st NN
3. In this z slice, find 4 NNs. This is the 1st set.
4. Remove all points in the z slice of the 1st NN from future searches
5. In the remainder of the cloud, find 2nd NN
6. Create a second z slice for all points matching z of 2nd NN
7. In this z slice, find 4 NNs. This is the 2nd set.
8. Return the 2 sets of 4 NNs, each set being on a different z slice.

## To test

You should see output similar to this at startup:
```
$ ign gazebo -v 4 buoyant_tethys.sdf
...
[Dbg] [ScienceSensorsSystem.cc:891] Number of points matching z == -0: 18471
[Dbg] [ScienceSensorsSystem.cc:947] Found 4 neighbors in this slice.
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (-8.3819e-13, -4.65661e-14, -0), distance 1.01812e-05 m
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (-0.0954054, -4.65661e-14, -0), distance 0.0954054 m
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (0.0954592, -4.65661e-14, -0), distance 0.0954592 m
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (-0.0142803, -0.180422, -0), distance 0.180986 m
[Dbg] [ScienceSensorsSystem.cc:850] Slice 2 nnIdx 4021, dist 4.99999
[Dbg] [ScienceSensorsSystem.cc:891] Number of points matching z == -5: 18471
[Dbg] [ScienceSensorsSystem.cc:947] Found 4 neighbors in this slice.
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (-8.3819e-13, -4.65661e-14, -5), distance 4.99999 m
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (-0.0954054, -4.65661e-14, -5), distance 5.0009 m
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (0.0954592, -4.65661e-14, -5), distance 5.0009 m
[Dbg] [ScienceSensorsSystem.cc:955] Neighbor at (-0.0142803, -0.180422, -5), distance 5.00326 m
```
Significance is that the 2 slices need to have different z's. Otherwise they're useless.

You should also verify that science data visualization still works - follow steps in #70.
The reason is that, if `FindInterpolators()` is called too often (every `PostUpdate()` multiple times), visualization never shows up.
I added a threshold `INTERPOLATE_DIST_THRESH` of 50 meters for the robot to move, before that function is called.
That threshold will need to be decrease when we actually test, because we currently downscale the data by a LOT to work around the orbiting tools.

## Future work

To test these thoroughly, we still need

- Actual trilinear interpolation, in a followup PR
- There needs to be a check for a corner case where the robot is at the boundary of a cluster of data, and its two closest z slices are both below (or both above) the robot, instead of one above and below. The next one above might not be for a very long way. The current algorithm would return both z slices below, incorrect for interpolation.

Then we can test statically and/or dynamically, which need
- Statically by spawning vehicle at different lat/long. Prerequisite: All science data need to be shifted to the new lat/long that the world origin is associated with, when vehicles are spawned using `empty_environment.sdf`. Currently, this is not implemented yet (see stub function `ShiftDataToNewSphericalOrigin()`.
- (Might be able to test statically by hardcoding a different vehicle XYZ in the SDF?)
- Dynamically. Move vehicle enough distance to test sensor data changing (acceptance mission probably does this for us)


lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -56,12 +61,75 @@ class tethys::ScienceSensorsSystemPrivate
...
+  /// Comparison between points is arbitrary. This function is only used for
+  /// set operations, not for literal sorting.
+  public: bool comparePclPoints(
+    const pcl::PointXYZ &a,
+    const pcl::PointXYZ &b);
+
+  /// \brief Find XYZ locations of points in the two closest z slices to
+  /// interpolate among.
+  /// \param[in] _pt Location in space to interpolate for
+  /// \param[in] _inds Indices in _arr
```
tfoote, line 83: _arr is undefiend in this scope

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -612,39 +701,304 @@ void ScienceSensorsSystemPrivate::GenerateOctrees()
...
+  const pcl::PointXYZ &a,
+  const pcl::PointXYZ &b)
 {
-  this->cloudPub.Publish(this->PointCloudMsg());
-  this->tempPub.Publish(this->tempMsg);
-  this->chlorPub.Publish(this->chlorMsg);
-  this->salPub.Publish(this->salMsg);
+  // Comparison between points is arbitrary. This function is only used for
+  // set operations, not for literal sorting.
+  if (a.x < b.x)
```
tfoote, line 721: This could be done in one line 
```suggestion
  return a.x < b.x && a.y < b.y && a.z < b.z
```

removing the rest of the lines.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -612,39 +701,304 @@ void ScienceSensorsSystemPrivate::GenerateOctrees()
...
+  std::vector<int> &_zSliceInds,
+  std::vector<int> &_interpolatorInds,
+  std::vector<float> &_interpolatorSqrDists,
+  std::vector<pcl::PointXYZ> &_interpolators,
+  int _k)
+{
+  // Find indices of points with z matching the given z. These points are on
+  // the same z slice. Logical indexing, produces Boolean array.
+  // Take only the block of the matrix containing the z values of the points.
+  Eigen::Matrix<bool, 1, Eigen::Dynamic> depthInds =
```
tfoote, line 912: Floating point equality comparisions are very fragile. We shouldn't rely on them here. 

There's PCL functionality that I think will do this much more cleanly: https://pcl.readthedocs.io/projects/tutorials/en/latest/passthrough.html#passthrough and hopefully efficiently.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -612,39 +701,304 @@ void ScienceSensorsSystemPrivate::GenerateOctrees()
...
+    interpolatorInds2, interpolatorSqrDists2, _interpolators2);
+  if (interpolatorInds2.size() < 4 || interpolatorSqrDists2.size() < 4)
+  {
+    ignwarn << "Could not find enough neighbors in 2nd slice z = " << nnZ
+      << " for trilinear interpolation." << std::endl;
+  }
+}
+
+/////////////////////////////////////////////////
+void ScienceSensorsSystemPrivate::SearchInDepthSlice(
```
tfoote, line 898: I would suggest separating the slice generation from the "SearchInSlice" method. 

The slices can be generated separately prior to running the search and then passed into the search function instead of returning the sliced cloud from the first search and generating a new slice for the second search.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -612,39 +701,304 @@ void ScienceSensorsSystemPrivate::GenerateOctrees()
...
+        _points(1, i),
+        _points(2, i));
+    }
+  }
+
+  // Make sure number of points in z-slice cloud == number of pts matching z
+  assert(zSliceIdx == depthInds.count());
+
+  // Create octree for this depth slice
+  auto octree = pcl::octree::OctreePointCloudSearch<pcl::PointXYZ>(
```
tfoote, line 947: This could be pulled out too and just pass in the octree to keep the complexity of this function minimal and focused on doing the search.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -816,28 +1170,56 @@ void ScienceSensorsSystem::PostUpdate(const ignition::gazebo::UpdateInfo &_info,
...
-      // kNN search (alternatives are voxel search and radius search. kNN
-      // search is good for variable resolution when the distance to the next
-      // neighbor is unknown).
-      int k = 8;
+      // Don't need to interpolate EVERY PostUpdate(). That's overkill.
+      // Only need to do it after robot has moved a distance from when we did
+      // the previous interpolation
+      if (sensorLatLon.value().Distance(
+          this->dataPtr->lastSensorUpdateCartesian) <
+        this->dataPtr->INTERPOLATE_DIST_THRESH)
```
tfoote, line 1179: We shouldn't use this magic number. It should be related to the data density. The optimal would be to compute the bounding box of when a nearest neighbor will transition. However as a proxy some fraction of the distance to the nearest neighbor might make sense as a heuristic causing over computation but it will be data driven instead of a magic number.

mabelzhang: This is just the quickest way to cut down the number of times interpolation is run, so that it doesn't run every PostUpdate(). This code is not tested until we have interpolation, so this is more a preventative measure.

We can add optimization later, if speed is a problem. That requires some thought, because the data has variable density. That will be a PR in itself.

mabelzhang: I should add that the only reason this was introduced right now, is that none of the visualization was showing when interpolation was always called.

mabelzhang:
I just merged from `main`. We made a lot of changes in 2 other PRs, so it was more like a rebase. Git auto-merge made some mistakes.
There were a lot of conflicts because of indentation diffs from #98 interjected in between code blocks added to PostUpdate() in this PR. Hopefully there were no merge errors.

Coordinate system has been changed in other PRs. So the sample output looks different now. Still just paying attention to depths being 0 and 5 though:
```
[Dbg] [ScienceSensorsSystem.cc:583] At time slice 0, populated 92355 spatial coordinates.
[Dbg] [ScienceSensorsSystem.cc:1080] Searching around sensor Cartesian location 8.83035e-39, -2.65023e-13, -1.01813e-05
[Dbg] [ScienceSensorsSystem.cc:691] 92355 points
[Dbg] [ScienceSensorsSystem.cc:693] Slice 1 nnIdx 4021, dist 1.01813e-05
[Dbg] [ScienceSensorsSystem.cc:807] Number of points matching z == -0: 18471
[Dbg] [ScienceSensorsSystem.cc:856] Found 4 neighbors in this slice.
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (1.39148e-12, -1.1001e-13, -0), distance 1.01813e-05 m
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (-0.181203, 1.84071e-05, -0), distance 0.181203 m
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (0.181203, 1.84071e-05, -0), distance 0.181203 m
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (1.37634e-12, -0.221866, -0), distance 0.221866 m
[Dbg] [ScienceSensorsSystem.cc:766] Slice 2 nnIdx 4021, dist 4.99999
[Dbg] [ScienceSensorsSystem.cc:807] Number of points matching z == -5: 18471
[Dbg] [ScienceSensorsSystem.cc:856] Found 4 neighbors in this slice.
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (1.39148e-12, -1.1001e-13, -5), distance 4.99999 m
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (-0.181203, 1.84071e-05, -5), distance 5.00327 m
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (0.181203, 1.84071e-05, -5), distance 5.00327 m
[Dbg] [ScienceSensorsSystem.cc:864] Neighbor at (1.37634e-12, -0.221866, -5), distance 5.00491 m
```

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: e2efd70

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: e2efd70

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: 7c85554 refactors to use a PassThrough filter.

Here I did a `==` for speedy implementation, because the data we have is very clean with integers. The quick alternative is to instead do `abs(a-b) < 1e-6`, which is pretty standard.

PassThrough filter still requires specifying a min and max range, which means we still need to do the same thing - pass in thresholds like `1e-6`.

In its defense, the code here actually reimplements PassThrough filter at a lower level, with the added flexibility that we can give it any if-condition we want, not limited to a min/max range in a specified field. In the new code, using the filter allows for simpler code, but loses that low-level control.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: Part of 7c85554

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: Part of 7c85554

mabelzhang:
Ready for another pass.

This is one of the most time-sensitive pieces for milestone 4. I would appreciate help to get these interpolation PRs in as ASAP, so that there are fewer conflicts to resolve.

In the interest of time, we should focus on critical functionality, and leave less critical improvements to a future PR, opening issues if necessary. Reminder that code in here is by no means finalized. Multiple steps are still to come. This is the 1st of >=2 PRs for interpolation, to keep PRs small and easier to review. Some of the API here are already changed in my new branch to be the 2nd PR. So we're not looking for perfection here, just for algorithmic correctness.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -713,25 +1009,41 @@ void ScienceSensorsSystem::PostUpdate(const ignition::gazebo::UpdateInfo &_info,
         sensorPosENU.Y(),
         sensorPosENU.Z());
 
-    // kNN search (alternatives are voxel search and radius search. kNN
-    // search is good for variable resolution when the distance to the next
-    // neighbor is unknown).
-    int k = 8;
+    // Visualization never shows up when interpolation is always called.
+    // Quick fix: Don't need to interpolate EVERY PostUpdate(). That's overkill.
```
tfoote, line 1013:
```suggestion
    // Quick fix: Don't need to interpolate EVERY PostUpdate(). That's overkill.
    // TODO(tfoote) update this threshold to be data dependent not hard coded with INTERPOLATE_DIST_THRESH
```

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -101,13 +165,39 @@ class tethys::ScienceSensorsSystemPrivate
...
+  /// \brief Coordinates where sensor location was last interpolated.
+  /// Helps to determine whether sensor location needs to be updated
+  public: ignition::math::Vector3d lastSensorPosENU = {
+    std::numeric_limits<double>::max(),
+    std::numeric_limits<double>::max(),
+    std::numeric_limits<double>::max()};
+
+  /// \brief Distance robot needs to move before another data interpolation
+  /// (based on sensor location) takes place.
+  public: const float INTERPOLATE_DIST_THRESH = 5.0;
```
tfoote, line 184:
```suggestion
  // TODO(tfoote) Avoid this magic number see below for more context.
  private: const float INTERPOLATE_DIST_THRESH = 5.0;
```

Since this is a magic number that's unadjustable, lets suppress it away from the public API.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -124,6 +214,16 @@ class tethys::ScienceSensorsSystemPrivate
   /// in the ENU world frame.
   public: std::vector<pcl::PointCloud<pcl::PointXYZ>::Ptr> timeSpaceCoords;
 
+  /// \brief Resolution of spatial coordinates in meters in octree, for data
+  /// search.
+  public: float spatialRes = 0.1f;
```
tfoote, line 219:
```suggestion
  // TODO(tfoote) Magic Number that should be parameterized.
  public: float spatialRes = 0.1f;
```

mabelzhang:
Might close if #110 is a good alternative.
On hold until that is clear after thorough tests some time later. Not urgent anymore for Q4.

tfoote (commented):
I see you commented that this is on hold pending the replacement in #110 which has an alternative implementation. I'll focus over there.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: Just a note that this is Pimplized. It is a private class, so everything in the class is already private. Just Ignition standard style - private class, all members are public.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: Fixed in #110 and now merged into this PR in 0af6a57

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: Fixed in #110 and now merged into this PR in 0af6a57

mabelzhang:
In the meeting today, we verbally agreed that this is still needed, because I remembered why we needed trilinear interpolation in the first place while debugging in https://github.com/osrf/lrauv/pull/126#pullrequestreview-838162680

I have a hybrid approach using both methods that might work. That is in progress in the `mabelzhang/trilinear_interp` branch that's to come after this PR.

In this PR, I merged from `main` in 0af6a57. Since `main` already has barycentric interpolation, there were a bunch of conflicts and a big diff. some of the function signatures here had to be changed. There's a new Boolean to choose between barycentric and trilinear interpolations. The dummy `TrilinearInterpolate()` that just returns the closest point had to be updated to work with the new API. It doesn't matter too much - it's already replaced in `mabelzhang/trilinear_interp` with the actual interpolation.

tfoote (commented):
I have a few changes and comments inline here. And the search algorithm I wanted to suggest refactoring it to simply do two slices and then find the nearest neighbors. I also found some magic numbers and removed some unnecessary arguments. 

I've put it forth in #142 

Also looking at the CI logs we're generating a lot of debug info when the data isn't present. We might want to make things less verbose, and not try as hard if there's no data in the world.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -606,6 +710,249 @@ void ScienceSensorsSystemPrivate::PclVectorToEigen(
...
+  bool _invert)
+{
+  // Separate a z slice, i.e. points with z equal to that of 1st NN
+  // Pass in extract_removed_indices=true to get indices of removed points
+  pcl::PassThrough<pcl::PointXYZ> passThruFilter =
+    pcl::PassThrough<pcl::PointXYZ>(true);
+  passThruFilter.setInputCloud(_cloud.makeShared());
+  passThruFilter.setNegative(_invert);
+  passThruFilter.setFilterFieldName("z");
+  passThruFilter.setFilterLimits(_depth - 1e-6, _depth + 1e-6);
```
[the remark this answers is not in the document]
mabelzhang: `1e-6` is actually a very standard tolerance used in a lot of places in Ignition, to do an `a == b` equivalence check while catching floating point errors. Lines 859-860 essentially check for `z == _depth`.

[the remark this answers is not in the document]
tfoote: There's a potential debate about what the range should be. But it definitely should not be a magic number in the code here. It should be passed in at a higher level such that it has potential to be adjusted for the different dataset if required.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -53,13 +54,100 @@ class tethys::ScienceSensorsSystemPrivate
...
+
+  /// \brief Find XYZ locations of points in the two closest z slices to
+  /// interpolate among.
+  /// \param[in] _pt Location in space to interpolate for
+  /// \param[in] _inds Indices of nearest neighbors to _pt
+  /// \param[in] _sqrDists Distances of nearest neighbors to _pt
+  /// \param[out] _interpolators1 XYZ points on a z slice to interpolate among
+  /// \param[out] _interpolators2 XYZ points on a second z slice to interpolate
+  /// among
+  public: void FindTrilinearInterpolators(
```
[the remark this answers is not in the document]
mabelzhang: Yeah, they should be cleared. feb33ed
We don't have standard success codes, at least in this project. I haven't seen them much in ign-gazebo plugins, because most plugin functions are in a private class internal to individual plugins. They may be called from only one place ever (as is the case here), so it doesn't need to be so formal.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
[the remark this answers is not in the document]
mabelzhang: I see that you put an alternative in #142. Can we break just the part about this comment off into its own PR? See https://github.com/osrf/lrauv/pull/142#issuecomment-1021920916 I would really like to base proposals for big changes on #127, where the interpolation algorithms are implemented, because even if the search in #83 gives us some results, we now know that it doesn't mean those neighbors will work with interpolation, because of the constraints and edge cases in the interpolation algorithms.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -606,6 +710,249 @@ void ScienceSensorsSystemPrivate::PclVectorToEigen(
...
+  }
+
+  // Step 2: exclude z slice of 1st NN from further searches.
+
+  // Remove all points in the z slice of the 1st NN, so that the 2nd NN will be
+  // found in another z slice.
+  // Set invert flag to get all but the depth slice.
+  pcl::PointCloud<pcl::PointXYZ> cloudExceptZSlice1;
+  std::vector<int> indsExceptZSlice1;
+  this->CreateDepthSlice(nnZ, *(this->timeSpaceCoords[this->timeIdx]),
```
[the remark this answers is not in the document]
mabelzhang: That saves performance, but it does add more code. f3f72c2

We need a `ExtractIndices` filter to get the inverted sub-cloud. It's less clean to read, because [ExtractIndices::setIndices()](http://docs.ros.org/en/hydro/api/pcl/html/classpcl_1_1ExtractIndices.html) only accepts PCL `IndicesPtr` type. It doesn't accept the `std::vector` returned from [PassThrough](http://docs.ros.org/en/hydro/api/pcl/html/classpcl_1_1PassThrough.html). So we have to convert the type first.

I tested the new replacement with the 3 csv files in `buoyant_tethys.sdf` and verified the output looks the same before and after this change. But again, the outputs from this PR are not correct, so I would not trust things beyond small mechanical changes.

mabelzhang:
I addressed the comments individually here. Please see my comment here https://github.com/osrf/lrauv/pull/142#issuecomment-1021920916 about moving big changes to the review process of #127.

The `DEBUG_INTERPOLATE` flag has been wrapped around all debug messages in #127.

tfoote (approved):
Sure we can merge this as is now and then target the suggestions from #142 onto #127 instead of this one.

PR #85, opened 2021-11-18
Suggestions to #70

chapulina:
In the interest of eventually upstreaming the visualization plugin to `ign-gazebo` (i.e. https://github.com/ignitionrobotics/ign-gazebo/issues/1156), I'm making some suggestions to keep the plugin generic. This also means it will automatically scale to other science data sensors that we add in the future.

viz_pc

* Instead of hardcoding all `FloatV` topics, populate the dropdown with all topics publishing that message type
* Take min / max value from the current float data (this may be inefficient, and potentially change the color for the same sample with each new message)
* Made the colors configurable from the GUI

These are just some initial improvements to make sure we're not locking the plugin into a single use case. There are many more usability and functionality fixes that we need to make before actually upstreaming this.

### DIFF
--- a/lrauv_ignition_plugins/src/VisualizePointCloud.cc
+++ b/lrauv_ignition_plugins/src/VisualizePointCloud.cc
@@ -33,8 +33,11 @@
 
 #include <ignition/plugin/Register.hh>
 
+#include <ignition/math/Color.hh>
 #include <ignition/math/Vector3.hh>
 #include <ignition/math/Pose3.hh>
+
+#include <ignition/msgs/Utility.hh>
 
 #include <ignition/transport/Node.hh>
 
@@ -110,6 +113,18 @@ namespace tethys
 
     /// \brief Parameter to calculate Marker size in z.
     public: const float RES_Z = 10;
+
+    /// \brief Minimum value in latest float vector
+    public: float minFloatV{std::numeric_limits<float>::max()};
+
+    /// \brief Maximum value in latest float vector
+    public: float maxFloatV{std::numeric_limits<float>::min()};
+
+    /// \brief Color for minimum value
+    public: ignition::math::Color minColor{255, 0, 0, 255};
+
+    /// \brief Color for maximum value
+    public: ignition::math::Color maxColor{0, 255, 0, 255};
   };
 }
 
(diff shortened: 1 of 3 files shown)


lrauv_ignition_plugins/src/VisualizePointCloud.cc
```diff
@@ -51,32 +54,26 @@ namespace tethys
...
-    public: ignition::msgs::Float_V tempMsg;
+    public: ignition::msgs::PointCloudPacked pointCloudMsg;
 
-    /// \brief Chlorophyll data to visualize
-    public: ignition::msgs::Float_V chlorMsg;
-
-    /// \brief Salinity data to visualize
-    public: ignition::msgs::Float_V salMsg;
+    /// \brief Message holding a float vector.
+    public: ignition::msgs::Float_V floatVMsg;
```
mabelzhang, line 76: I think we should let the user have the capability of visualizing more than one type of data at a time, because sometimes it's useful to overlap multiple modalities and see how they relate. For example, temperature and salinity together directly imply density, while either one of them alone doesn't.

We might add the capability of adding more topic drop-down lists, rather than restricting to subscribing only to one topic at a time.

mabelzhang: If we don't want to hardcode them, they can be in a vector that expands and contracts. I can open an issue about that.

mabelzhang (approved):
Nice! I tested it and it works as pictured.

I left one comment about visualizing multiple topics above. Since the current version does not disable anything in #70, as neither have the GUI ability to add multiple topics, I'm going to approve so that I can rebase #85 on these changes.

But we should keep in mind that we should extend `floatVMsg` to an array of different topics, and allow subscribing to multiple types, so not necessarily unsubscribing to all other topics right away when a new topic is selected.

For example, one might add salinity as translucent light blue to dark blue, and temperature as translucent light red to dark red. Then by looking at what's dark purple, you'll be able to tell what areas have the highest density. Once we have ocean current data, visualizing multiple topics would be even more relevant, because e.g. 3D current flow is dependent on density, and other such relationships.

The most intuitive way is probably to have a + button to add more rows of topics, and let the user choose different colors. Pretty much like how RViz handles multiple topics of markers.

Let me know if you want to make further changes to this PR. I'll let you merge.

lrauv_ignition_plugins/src/VisualizePointCloud.cc
mabelzhang: Speaking of not hardcoding, I saw there was negative data in chlorophyll and wasn't sure if it's an error or intended. So I capped off at 0 and visualized everything below 0 at the same color as 0. Looks like in this subset of data we're visualizing, we don't have negatives. Just curious @braanan are those values intended to be negative?

chapulina: > we should let the user have the capability of visualizing more than one type of data at a time

Oh sorry, I was going to add a gif but forgot. That use case is already supported, here it goes:

multiple_data

mabelzhang: Awesome!

Issue #54, opened 2021-10-28
Bump to Garden 🌱


chapulina:
Moving https://github.com/osrf/lrauv_private/pull/147 to an issue:

---

@arjo129 :

This is actually quite non-trivial it will require the following:
* [x] Fixes on the CI since CI pulls fortress nightlies
* [ ] Rebuilding the MBARI docker image and nesscary code changes on the MBARI side.
* [x] Bumping version numbers in CMakeFiles

Moving to garden will allow us to make breaking changes to Ignition without any problems.

---

Currently Garden still matches Fortress, as we haven't officially decided to bump any libraries yet. 

I believe this project may need bumps for:

* `ign-msgs` and downstream libs: We may be adding fields to the marker and/or wrench messages (#148 )

PR #86, opened 2021-11-19
Remove pre-release repos, Fortress is stable

chapulina:
No need for the pre-release repo anymore. This shouldn't really be changing anything at this point, but it's safer.

Once we go to Garden (#54 ), we should add the nightly repo.


arjo129:
Will this mean that if we upstream a bugfix/feature, there will need to be a new minor release every time?

chapulina:
> Will this mean that if we upstream a bugfix/feature, there will need to be a new minor release every time?

Yeah, we only build nightlies for the unstable version, which is Garden now. Pre-releases and stable releases need to be manually triggered.

Once we go to Garden we should use nightlies again. I just wanted to remove this so there's no impression that we're actually using any pre-releases, because we aren't.

tfoote (approved):
This looks fine as is. 

I was going to ask what the harm is of having the prereleases available if it's not going to be updated until we go to garden. Thus if we're planning to go to garden in the near future we could just keep it there. 

But then I'd like to refactor this into a multi-stage build or parameterize this so cleaning this out for now makes sense either way..

PR #87, opened 2021-11-20
Setup profiler

chapulina:
[Ignition Common's profiler](https://ignitionrobotics.org/api/common/4.4/profiler.html) is very handy to debug performance issues. I'm planning to use it to look into the performance of the markers.

This PR just sets up the profiler and adds some profiling points. These points will change with PRs like #70, so don't worry too much about them right now, they're just placeholders.

Example output:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ Clone this repository then run
 colcon build
 ```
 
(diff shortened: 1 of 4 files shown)


caguero (approved):
Works as expected:
lrauv_profiler

PR #90, opened 2021-11-23
🤦‍♀️ Fix thruster coefficient

chapulina:
From #76:

> The formula on TethysComm was using 0.0016 m for the propeller diameter, but the Thruster uses 0.2 m, taken from the SDF.

Well... I forgot the `^4` ... `0.0016 == 0.2 ^ 4` :grimacing: This PR reverts it back.

>  I updated TethysComm to use 0.2 m, and that made the vehicle extremely fast. So I reduced the coefficient 100x to compensate.

This PR reverts it back... :grimacing: 

> Updated a few test expectations to the new thrust.

Yeah, updating that again...

> we're converting from ang vel to force and back. It would reduce errors if we only converted once.

Case in point :upside_down_face: 

---

I discovered this while working on a test for the state message on #81 . I broke off just the `propOmega` checks into this PR.

### DIFF
--- a/lrauv_description/models/tethys_equipped/model.sdf
+++ b/lrauv_description/models/tethys_equipped/model.sdf
@@ -94,7 +94,7 @@
         <namespace>tethys</namespace>
         <joint_name>propeller_joint</joint_name>
         <!-- Be sure to update TethysComm when updating these numbers -->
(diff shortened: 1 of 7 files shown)


arjo129 (approved):
Thanks  for spotting this.

PR #91, opened 2021-11-23
Don't publish LRAUV state while paused

chapulina:
`TethysComm` has been publishing the latest state even while simulation is paused. 

Not publishing while paused helps with debugging, because if we're echoing messages or checking console logs, we can step through them and analyze the latest data slowly. This may also help the controller, I'm not sure what it's doing with all these repeated messages that it is receiving.

One downside of this is that some sensor data may be received between the last publish and the pause, and it won't be published until unpause. But I'm not sure this causes any actual issues.

---

I also went ahead and deleted an unused variable, and moved another one down closer to where it is used.

### DIFF
--- a/lrauv_ignition_plugins/src/TethysCommPlugin.cc
+++ b/lrauv_ignition_plugins/src/TethysCommPlugin.cc
@@ -484,8 +484,8 @@ void TethysCommPlugin::PostUpdate(
   const ignition::gazebo::UpdateInfo &_info,
   const ignition::gazebo::EntityComponentManager &_ecm)
 {
-  ignition::gazebo::Link baseLink(this->modelLink);
-  auto modelPose = ignition::gazebo::worldPose(this->modelLink, _ecm);
+  if (_info.paused)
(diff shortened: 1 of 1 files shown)


arjo129 (approved):
Thanks, tested on my computer against the LRAUV-application and it works. This is a great QOL improvement.

mabelzhang:
The only thing I can think of is the time syncing on the MBARI side. That is done with the `/stats` message now, so I THINK it is okay. (Two timestamps are being written to the Slate, and I don't know how often that's written. One is the Ignition time from the State message, and the other is a `timeExtWriter_` that writes the elapsed time. I don't remember whether that picks up the pause from a higher controller.)

We made particular effort to make sure time sync was still in perfect step even after pausing for a long time.

To be prudent, someone should test the yoyo mission manually while pausing it a few times in the middle, and make sure it doesn't make sudden turns or weird movements.

arjo129:
Yea I tested that behavior seems to work as expected.

PR #101, opened 2021-11-30
Improvements to documentation and docker setups from my getting spun up

tfoote:
This is an aggregate of my making progress on learning how things are working and trying to streamline things a bit. 

The usage of rocker requires a new feature I just added to work around the upstream image setup: https://github.com/osrf/rocker/pull/164


docker/tests/Dockerfile
```diff
@@ -67,9 +66,12 @@ RUN rm -rf build log install src/lrauv
...
 RUN [ "/bin/bash" , "-c" , "colcon build --cmake-args='-DBUILD_TESTING=true'" ]
 
+COPY docker/entrypoint.bash /
+# setup ignition environment
+ENTRYPOINT ["/entrypoint.bash"]
+
 # Run tests
-RUN [ "/bin/bash" , "-c" , ". ~/lrauv_ws/install/setup.sh; colcon test --event-handlers console_direct+" ]
-RUN [ "/bin/bash" , "-c" , ". ~/lrauv_ws/install/setup.sh; colcon test-result" ]
+CMD [ "/bin/bash" , "-c" , "colcon test --event-handlers console_direct+; colcon test-result" ]
```
chapulina, line 77: I don't see the test output on GitHub actions, did it run and the output got swallowed?

chapulina (commented):
> The usage of rocker requires a new feature I just added 

Do we need to wait for a release before merging this PR?

docker/tests/Dockerfile
tfoote: Ahh, the actions aren't actually running the image, only building the image. I'll add it to the action.

docker/build_and_run_docker.sh
```diff
@@ -23,55 +23,11 @@
...
-fi
-
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 cd $DIR
 # Build the docker image
-docker build -t mbari_lrauv -f $DIR/empty_world/Dockerfile ..
+docker build -t mbari_lrauv -f $DIR/tests/Dockerfile ..
+
+rocker --nvidia --x11 --user --user-override-name=developer --user-preserve-home --  mbari_lrauv
```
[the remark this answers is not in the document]
tfoote: Good point. I've homogenized the rocker and docker content.

tfoote:
I have released rocker 0.2.7 to get the new command line argument

README.md
```diff
@@ -24,7 +24,9 @@ Optionally, you may choose to build this repository using Docker, for
 convenience.
 Make sure you have a recent version of [Docker](https://docs.docker.com/) and
 [nvidia-docker](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#docker)
-installed. Next to get started simply run the following command.
+installed.
+And [rocker](https://github.com/osrf/rocker) available as debian package `python3-rocker` with the ROS packages.
```
chapulina, line 28: Nitpick: this project doesn't use ROS, so users landing here may not have ROS packages setup. They may setup it at later stages to get `colcon`, but up here on the README, it may not be clear how to set it up.

I always prefer to err on the side of explicitness in this situation to avoid confusing new users who may not be familiar with ROS packages. So I'd suggest either removing the part about "_debian package `python3-rocker` with the ROS packages_" and letting the `rocker` entry page explain how to set it up, or repeating the step-by-step here on the README.

README.md
```diff
@@ -36,6 +38,41 @@ docker/join.sh mbari_lrauv
...
+
+~ 90 minutes from scratch
+
+~ 70 seconds with only workspace changes
+
+```
+cd docker
+docker build -t mbari_lrauv -f empty_world/Dockerfile ..
+rocker --nvidia --x11 --user mbari_lrauv bash
+ign launch lrauv_world.ign
```
chapulina, line 52: The launch file was removed in #102 , we should use plain `ign gazebo` commands now

docker/tests/Dockerfile
```diff
@@ -35,27 +35,26 @@ RUN apt-get update \
...
         gnupg2 \
+        libpcl-dev \
         lsb-release \
+        tmux \
+        tmuxinator \
         tzdata \
-        wget
+        wget \
+  && rm -rf /var/lib/apt/lists/* \
+  && apt-get -qq clean
```
chapulina, line 45: Is it worth it cleaning it here, considering we'll do an update below again?

chapulina (approved):
LGTM, I just have minor comments and I think we should wait for @arjo129 's approval too.

docker/debug_integration.sh
```diff
@@ -23,60 +23,14 @@
...
-  -v "/etc/localtime:/etc/localtime:ro" \
-  -v "/dev/input:/dev/input" \
-  -v "$DIR/../../results:/results"\
-  --network host \
-  --rm \
-  --privileged \
-  --security-opt seccomp=unconfined \
-  $DOCKER_OPTS \
-  mbari_lrauv_debug
+rocker --nvidia --x11 --user --user-override-name=developer --user-preserve-home -- mbari_lrauv_tests tmuxinator start debug_and_plot -n debug_session -p src/lrauv/docker/tests/debug_integration_mux.yml
```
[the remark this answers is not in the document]
arjo129: Good News: The system launches correctly with the correct panes.

Bad News: I may be doing something wrong but when I run via rocker I get the following error:
```
[GUI] [Err] [Ogre2RenderEngine.cc:905]  Unable to create the rendering window: OGRE EXCEPTION(3:RenderingAPIException): OpenGL 3.3 is not supported. Please update your graphics card drivers. in GL3PlusRenderSyste
m::initialiseContext at /var/lib/jenkins/workspace/ogre-2.2-debbuilder/repo/RenderSystems/GL3Plus/src/OgreGL3PlusRenderSystem.cpp (line 3364)                                                                       
[GUI] [Err] [Ogre2RenderEngine.cc:913] Unable to create the rendering window after [11] attempts.                                                                                                                   
[GUI] [Err] [Ogre2RenderEngine.cc:834] Failed to create dummy render window.
```

Might be some (nvidia?) driver issue or a problem with the way I setup rocker :thinking: . 
Also is the `--nvidia` really nessecary? What about scenarios where we aren't using an nvidia card. I don't think we really need cuda acceleration.

arjo129 (changes requested):
Thanks for these changes.

I've been unable to get rocker working on my comp correctly. Could be some GPU setup issue. I'm worrying if we really need the `--nvidia` flag. There are some minor knits in regards to style.

tfoote:
>Is it worth it cleaning it here, considering we'll do an update below again?

Yes, the standard practice when building dockerfiles is to clear the cache at every layer. This has two major reasons. First is that it keeps the image size down. Every file change in the system at the end of a step is saved into the layer. So if you do 3 apt-get updates in three consecutive layers you will get three copies of the cache in each layer assuming they're slightly different because they were pulled at different times. And secondly, because docker caches previous layers, you have no knowledge of how long ago the last update was so you are going to have to do another update. And the cost of the incremental update versus a fresh update is small, and gives you a much more consistent result.

mabelzhang:
Butting in with a nitpick request in my head for a while - can we rename this image so that it doesn't start with `mbari`? Technically, this is not mbari's image, so it might be called `osrf_lrauv` or something other than MBARI LRAUV.

Confusion arises because we host a different, official image on MBARI GitHub that has the MBARI code in it: https://hub.docker.com/r/mbari/lrauv-ignition-sim When you pull it, it's called `mbari/lrauv-ignition-sim`. At the end of the day, I end up with a handful of images under my `docker images`, all called some combination of MBARI LRAUV (the official no-MBARI-source one which I push to DockerHub, the with-source one that's a dependency, and my personal development one. Not counting this one in the osrf repo).

The more we can disambiguate these names, the less confusing it will be for ourselves.

tfoote:
Good idea @mabelzhang I've updated it with the osrf_lrauv to mirror the repo name for clarity.

tfoote:
I worked with @arjo129 on slack and we isolated his crashing to be related to the integrated AMD card getting selected and his `prime-select` settings seemingly not going into the environment.

chapulina (changes requested):

tfoote:
I've rebased this for garden. 

With the new rebuilds of the garden image with different UIDs it needs https://github.com/osrf/rocker/pull/168

docker/tests/Dockerfile
tfoote: Yes, this removes the contents of the apt cache from the layer, making the docker container much slimmer. Otherwise you just end up with outdated caches in every layer of the container.

hidmic:
> I worked with @arjo129 on slack and we isolated his crashing to be related to the integrated AMD card getting selected and his `prime-select` settings seemingly not going into the environment.

Circling back here. I did not find anything particularly unusual in `rocker`. @arjo129 I'll assume that you've `prime-select`'ed `nvidia`. Could it be an issue with your `LD_LIBRARY_PATH` (https://github.com/NVIDIA/nvidia-docker/issues/612)? Would you mind giving it a shot?

caguero:
Closing this PR for now.

PR #103, opened 2021-12-01
Salt water usage

arjo129:
* Builds on #96 and  #89

I'm opening this for visibility but @benranaan pointed out that the hydrodynamic parameters we are using are meant for salt water with a density of 1025KG/m^3 instead of the default 1000KG we have been using. This actually helps damp some oscillations without needing to tweak the hydrodynamic parameters.


arjo129:
Removing this PR due to its clobbered history

PR #104, opened 2021-12-01
Axis visualization: ENU, NED, FSK

chapulina:
Requires

* https://github.com/ignitionrobotics/ign-gui/pull/325

---

It can be tough to keep track of all the coordinate frames that we're dealing with. This plugin should help debugging. See more information on where each of these frames is used on #81.

Unfortunately, the floating text doesn't work on Ogre2 yet, so to see it, you need to use Ogre1:

`ign gazebo buoyant_tethys.sdf -v 4 -r --render-engine=ogre`

axes_moving

Latest iteration:

yoyo_axes

## TODO

* See https://github.com/osrf/lrauv/issues/111


arjo129 (approved):
Looking awesome. Tested this and it works for me.

Issue #99, opened 2021-11-30
Integrate `testPitchAndDepthMassVBS.xml` with our CI


arjo129:
Currently we test all the missions except for `testPitchAndDepthMassVBS.xml`, but we should also integrate the `testPitchAndDepthMassVBS.xml` with our CI.

PR #105, opened 2021-12-01
Addresses #99. Adds the pitchmass + vbs depth.

arjo129:
So after after a short discussion with @braanan it seems that the normal mode of using VBS to sink does lead to certain behaviors such as spinning and reverse motion. Therefore, MBARI normally uses both the mass shifter and the buoyancy engine to get the vehicle to sink. MBARI has a mission which does this already, however we aren't actually running this in CI. This PR runs this in CI.

Note: As the model in the main repo is not actually up to the mark, it does not perform properly on this test. Hence the expectations have been kept large. However when combined with some of the changes ion #89 and #59 it is possible to narrow the expectations down.


lrauv_ignition_plugins/test/test_mission_pitch_and_depth_mass_vbs.cc
```diff
@@ -0,0 +1,113 @@
...
+    EXPECT_LT(pose.Pos().Z(), 0.1);
+    // FIXME(arjo): This should dive to a max of 10m I think
+    EXPECT_GT(pose.Pos().Z(), -21.5);
+
+    // Vehicle should exhibit minimal lateral translation.
+    EXPECT_NEAR(pose.Pos().X(), 0, 10); // FIXME(arjo): IMPORTANT!!
+    EXPECT_NEAR(pose.Pos().Y(), 0, 1e-1);
+
+    // Vehicle should hold a fixed pitch
+    // FIXME(arjo): Shouldnt be pitching this much
```
chapulina, line 89: Just looking at the mission file, it's not clear to me that this is the expectation. Just double-checking, is this mission indeed supposed to make the vehicle sink without any pitch?

chapulina (approved):
LGTM, nice idea with the oscillation check

lrauv_ignition_plugins/test/test_mission_pitch_and_depth_mass_vbs.cc
arjo129: I just confirmed with @braanan that is the expected behaviour.

PR #106, opened 2021-12-04
Cleanup old science values and improve initial subscription

chapulina:
This addresses this comment by @mabelzhang : https://github.com/osrf/lrauv/pull/93#issuecomment-985776670

* Needs this to reduce warnings, but not for functionality: https://github.com/ignitionrobotics/ign-gui/pull/326

switch_science

---

* Added services for all `FloatV`s, which were missing. This is needed to get the data on demand, since Ignition Transport doesn't support latching.
* Make the `FloatV` optional. This is not a use case here, but I think it's a nice-to-have when we upstream.
* Instead of skipping NaNs when displaying, send a `DELETE_MARKER` command, so that if the data changes from valid to `NaN` it disappears.
* Removed the `-` from the marker namespaces because it looked awkward when there's no float topic

### DIFF
--- a/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
+++ b/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
@@ -53,10 +53,25 @@ class tethys::ScienceSensorsSystemPrivate
   /// \brief Publish the latest point cloud
   public: void PublishData();
 
-  /// \brief Service callback for a point cloud with the latest science data.
+  /// \brief Service callback for a point cloud with the latest position data.
   /// \param[in] _res Point cloud to return
   /// \return True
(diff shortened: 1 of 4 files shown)


mabelzhang (commented):
Works as intended. I see what you mean by the warnings.
So I'm actually unclear on why they have to be deleted again. I thought `ClearMarkers()` is called in `OnFloatVTopic()`, so shouldn't all the points already get deleted whenever the topic is switched?

mabelzhang (approved):
Will let you merge since we have some specific order of merging at the moment.

chapulina:
> I thought ClearMarkers() is called in OnFloatVTopic(), so shouldn't all the points already get deleted whenever the topic is switched?

Good point, maybe it's a timing thing? I should dig more later. I'll merge this now to unblock other PRs.

PR #108, opened 2021-12-07
Update density value in tethys comm plugin.

arjo129:
The density value is used by the vertical controller in the LRAUV's controller. Because we have not been populating this field with a non-zero value, the Vertical Controller would crash. For more robustness, the controller itself should start of with a default density value. 1000kgm^3 could be a good value.

### DIFF
--- a/lrauv_ignition_plugins/src/TethysCommPlugin.hh
+++ b/lrauv_ignition_plugins/src/TethysCommPlugin.hh
@@ -182,6 +182,9 @@ namespace tethys
     /// Latest chlorophyll data received from sensor. NaN if not received.
     private: float latestChlorophyll{std::nanf("")};
 
+    /// Ocean Density
+    private: double oceanDensity{1000};
+
     /// Latest current data received from sensor. NaN if not received.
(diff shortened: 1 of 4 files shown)


lrauv_ignition_plugins/src/TethysCommPlugin.hh
```diff
@@ -182,6 +182,9 @@ namespace tethys
     /// Latest chlorophyll data received from sensor. NaN if not received.
     private: float latestChlorophyll{std::nanf("")};
 
+    /// Ocean Density
```
chapulina, line 185:
```suggestion
    /// Ocean Density in kg / m ^ 3
```

chapulina (approved):
LGTM with a couple of non-blocking comments:

* It would be good to document the field in `lrauv_state.proto`. On #81 I'm just writing `Not populated`. I think it would make sense to document it as part of this PR, since this is where we're starting to populate it.
* We could explicitly set the density on `tethys_equipped`, even if it's reusing the value. Then it will be easier to spot it and change it (see #103 )

tfoote (approved):

PR #109, opened 2021-12-08
Add water velocity.

arjo129:
This micro-PR adds water velocity as one of the parameters we care about.

### DIFF
--- a/lrauv_ignition_plugins/src/TethysCommPlugin.cc
+++ b/lrauv_ignition_plugins/src/TethysCommPlugin.cc
@@ -588,6 +588,7 @@ void TethysCommPlugin::PostUpdate(
   // rateUVW
   // TODO(arjo): include currents in water velocity?
   auto localVel = modelPose.Rot().Inverse() * veloGround;
+  //TODO(louise) check for translation/position effects
   ROSToFSK(localVel);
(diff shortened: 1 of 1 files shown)


chapulina (approved):
I'm also populating this field in #81 . My concern with the `ROSToFSK` function is that it's just flipping axes, but not taking the whole vehicle pose into account.

If we want to get something in quick just to make sure the field is populated, I'm ok with merging this PR, but heads up that the values being published here aren't correct.

lrauv_ignition_plugins/src/TethysCommPlugin.cc
```diff
@@ -586,7 +586,10 @@ void TethysCommPlugin::PostUpdate(
 
   // Water velocity
   // rateUVW
-  // TODO(anyone)
+  // TODO(arjo): include currents in water velocity?
+  auto localVel = modelPose.Rot().Inverse() * veloGround;
+  ROSToFSK(localVel);
```
tfoote, line 591:
```suggestion
  //TODO(louise) check for translation/position effects
  ROSToFSK(localVel);
```

tfoote: Add a comment to track Louise's question.

tfoote (approved):
add one more comment/todo to track louise's concern.

PR #112, opened 2021-12-09
Make PitchMass test easier to maintain

chapulina:
We lose some granularity, but in the end of the day, it doesn't matter how quickly the vehicle drifted, just knowing that it drifted is enough to know there's something wrong.

The end goal is to reduce all tolerances to acceptable values throughout the entire mission.

### DIFF
--- a/lrauv_description/models/tethys_equipped/model.sdf
+++ b/lrauv_description/models/tethys_equipped/model.sdf
@@ -153,6 +153,7 @@
         <command_topic>tethys/command_topic</command_topic>
         <state_topic>tethys/state_topic</state_topic>
         <debug_printout>0</debug_printout>
+        <ocean_density>1000</ocean_density>
       </plugin>
       <plugin
         filename="HydrodynamicsPlugin"
--- a/lrauv_ignition_plugins/CMakeLists.txt
+++ b/lrauv_ignition_plugins/CMakeLists.txt
@@ -152,7 +152,10 @@ add_lrauv_plugin(AcousticCommsPlugin
     lrauv_internal_comms
   PRIVATE_LINK_LIBS
     CommsSupport)
-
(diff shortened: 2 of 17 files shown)


arjo129:
Hey @chapulina I actually have written a new version of the test here: https://github.com/osrf/lrauv/blob/2f32e9fdc6e65f8a9a7a048c6e0b2db18e1d1449/lrauv_ignition_plugins/test/test_mission_pitch_mass.cc

I can cherry pick this out in a separate PR if thats what you think is the correct way... I think though we should focus on the garden bump so that the tests in that PR #89 get fixed.

chapulina:
I'm fine going with your test. If you open a PR with just that I think we can get it in faster than if it comes bundled with other changes.

arjo129 (approved):
I'm approving this but, the PR is still failing... I recommend not fussing about this test at all since the behavior is broken. Perhaps it makes sense for now to disable this test completely. I really feel like tweaking this test is going to be a lot of unnecessary effort for you.

chapulina:
>  Perhaps it makes sense for now to disable this test completely. I really feel like tweaking this test is going to be a lot of unnecessary effort for you.

Fair enough. I'll try to make it as loose as possible, and then we update it properly on #89.

Right now it looks like it failed because of the RDI though:

```
2021-12-14T04:34:39.746Z,1639456479.746 [MissionManager](ERROR): No startup, active, or default mission![0m
2021-12-14T04:34:40.165Z,1639456480.165 [MissionManager](ERROR): No startup, active, or default mission![0m
2021-12-14T04:37:52.505Z,1639456672.505 [RDI_Pathfinder](ERROR): Data Fault[0m
2021-12-14T04:37:52.505Z,1639456672.505 [CBIT](ERROR): Data Fault in component: RDI_Pathfinder[0m
```

Let me rerun it

PR #113, opened 2021-12-10
Use actual depth vbs expectations

arjo129:
This test is a rewrite of the testDepthVBS test. The key change is instead of a trajectory, we check that the vehicle eventually reaches a depth of 10m and its depth rate eventually reaches zero. It also checks that the velocity is reducing slowly. I expect this to be a :x: till the upstream changes in the controller are merged and a new docker image is generated,


lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -85,19 +85,37 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
-  this->CheckRange(14000, {10.69, 10.22, -16.95, -0.00, -0.03, -2.80}, {15.0, 3.14});
-  this->CheckRange(16000, {7.85, 8.85, -16.07, -0.00, -0.01, -2.13}, {15.0, 3.14});
-  this->CheckRange(18000, {6.48, 6.29, -13.84, 0.00, -0.07, -1.59}, {15.0, 3.14});
-  this->CheckRange(20000, {6.99, 2.26, -10.73, 0.00, -0.17, -0.93}, {15.0, 3.14});
-  this->CheckRange(22000, {11.22, -0.97, -7.71, -0.00, -0.10, 0.06}, {15.0, 3.14});
-  this->CheckRange(24000, {15.47, 0.32, -5.62, -0.00, 0.04, 0.98}, {15.0, 3.14});
-  this->CheckRange(26000, {16.95, 3.13, -5.16, 0.00, 0.00, 1.66}, {15.0, 3.14});
-  this->CheckRange(28000, {16.65, 5.57, -6.79, -0.00, 0.02, 2.17}, {15.0, 3.14});
+  ignition::math::Vector3d maxVel(0, 0, 0);
+  for (int i = 1; i <= this->tethysPoses.size(); i ++)
```
chapulina, line 89: How about not checking every single pose? It takes longer with little benefit.

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -85,19 +85,37 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
+  {
+    // Vehicle roll should be constant
+    EXPECT_NEAR(tethysPoses[i].Rot().Euler().X(), 0, 1e-2);
+
+    // Vehicle should not go vertical
+    EXPECT_LT(tethysPoses[i].Rot().Euler().Y(), IGN_DTOR(40));
```
chapulina, line 95: Should we check the other direction as well? As on `EXPECT_GT...`

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -85,19 +85,37 @@ TEST_F(LrauvTestFixture, DepthVBS)
   //       << std::endl;
   // }
```
chapulina, line 86: We can remove this block too

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -85,19 +85,37 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
+  for (int i = 1; i <= this->tethysPoses.size(); i ++)
+  {
+    // Vehicle roll should be constant
+    EXPECT_NEAR(tethysPoses[i].Rot().Euler().X(), 0, 1e-2);
+
+    // Vehicle should not go vertical
+    EXPECT_LT(tethysPoses[i].Rot().Euler().Y(), IGN_DTOR(40));
+
+    // Vehicle should not exceed 20m depth
+    EXPECT_GT(tethysPoses[i].Pos().Z(), -20);
```
chapulina, line 98: Looking at the mission XML, the depth command is 10 m with a deadband of 0.1 m.

* Can we document that here?
* Can we make the expectation a bit tighter?

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -85,19 +85,37 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
+    {
+      maxVel = tethysLinearVel[i];
+    }
+  }
+
+  // Vehicle's final pose should be near the 10m mark
+  EXPECT_NEAR(tethysPoses.back().Pos().Z(), -10, 1e-2);
+
+  // Vehicle's final pose should be near the 10m mark
+  EXPECT_NEAR(tethysPoses.back().Pos().Z(), -10, 1e-2);
```
chapulina, line 110: This is repeated

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -85,19 +85,37 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
+
+  // Vehicle should have almost zero Z velocity at the end.
+  EXPECT_NEAR(tethysLinearVel.back().Z(), 0, 1e-1);
+
+  // Expect velocity to approach zero
+  EXPECT_LT(tethysLinearVel.back().Length(), maxVel.Length());
```
chapulina, line 116: Shouldn't we be comparing to zero instead of `maxVel`?

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -85,19 +85,37 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
+
+  // Vehicle should pitch backward slightly
+  EXPECT_GE(tethysPoses.back().Rot().Euler().Y(), 0);
```
chapulina, line 119: If it's slightly, now about using `EXPECT_NEAR` with a small tolerance? With `GE` it could be really large and we wouldn't notice.

chapulina (commented):
Thanks for cleaning up the test!

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
arjo129: So while the depth is 10m, it is not possible to cleanly descend to 10m, the controller will overshoot to about 16m before coming back to 10m as shown in the graph below. I believe this is consistent with the tank tests. The idea here is to manage the overshoot. I'll update the comment so its clearer.

image

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
arjo129: ec1c2fd

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
arjo129: ec1c2fd

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
arjo129: Good Point, I'm adding an `EXPECT_LE` to the test in e9d5bc

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
chapulina: Got it, makes sense. Then can we document that in a comment here? For future us looking at the test.

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
arjo129: updated comment

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
chapulina: Ping @arjo129 , I think this comment is especially important now that the number of iterations was increased

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -71,34 +71,47 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
-  this->CheckRange(24000, {15.47, 0.32, -5.62, -0.00, 0.04, 0.98}, {15.0, 3.14});
-  this->CheckRange(26000, {16.95, 3.13, -5.16, 0.00, 0.00, 1.66}, {15.0, 3.14});
-  this->CheckRange(28000, {16.65, 5.57, -6.79, -0.00, 0.02, 2.17}, {15.0, 3.14});
+  int maxIterations{28000};
+  ASSERT_LT(maxIterations, this->tethysPoses.size());
+
+  ignition::math::Vector3d maxVel(0, 0, 0);
+  for (int i = 1; i <= this->tethysPoses.size(); i ++)
+  {
+    // Vehicle roll should be constant
```
chapulina, line 80: Should we check other quantities that should also be constant? i.e. X, Y and yaw?

chapulina (approved):
LGTM, I just have some recommendations below to make the test quicker and to check other directions (X, Y and yaw), so we can catch problems if we introduce (more) drift in the future.

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
arjo129: X, Y and Yaw are not constant. This was also found in real life data. There is a significant amount of translation along the X axis and yawing.

chapulina: Do we know how much change we should expect on these dimensions? Is there a point when it's too much?

arjo129:
I get the concern over excessive translation. We don't have such data. As I
said earlier it is ok for the vehicle to translate and spin (significantly
as normally the pitchmass is used to prevent the vehicle from excessive
pitching and thus translation). If you look at the end of the test we do
check that the overall vehicle velocity reduces. If we really want we can
run the test for longer till the velocity approaches zero. It doesn't make
sense for us to set upper bounds on translation as there is no clear amount
and we will have to keep changing it. The test pitchmass + depth VBS
mission however should have some bounds on translation.

chapulina:
> it is ok for the vehicle to translate and spin (significantly as normally the pitchmass is used to prevent the vehicle from excessive pitching and thus translation).
> The test pitchmass + depth VBS mission however should have some bounds on translation.

I see, ok, so it sounds like it's difficult (impossible?) to use this test to detect regressions on our dynamics if it introduces extra drifting, and we have other tests that can be used to check that instead :+1:

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
chapulina: We can address this later

chapulina:
I'll merge this so the test becomes stable for other branches

PR #114, opened 2021-12-10
Log LRAUV application errors as errors.

arjo129:
In many cases we find "flaky" mission integration tests. Often this is because of components crashing on the LRAUV controller side. This often leads us down a very painful debugging journey. This patch proposes the following change: if there is an error reported by the LRAUV application, report it. The hope is that this will save us valuable time when debugging.

### DIFF
--- a/lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
+++ b/lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
@@ -213,6 +213,14 @@ class LrauvTestFixture : public ::testing::Test
         // FIXME: LRAUV app hangs after quit, so force close it
         // See https://github.com/osrf/lrauv/issues/83
         std::string bufferStr{buffer};
+
+        std::string error{"ERROR"};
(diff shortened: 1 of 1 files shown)


chapulina (approved):
Sweet!

lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
```diff
@@ -213,6 +213,14 @@ class LrauvTestFixture : public ::testing::Test
         // FIXME: LRAUV app hangs after quit, so force close it
         // See https://github.com/osrf/lrauv/issues/83
         std::string bufferStr{buffer};
+
+        std::string error{"ERROR"};
```
braanan, line 217: Thanks, @arjo129! I think we really only care about CRITICAL failures, since those will abort the mission

chapulina: Here we go: https://github.com/osrf/lrauv/pull/120

PR #115, opened 2021-12-10
Temporary work around to ensure pathfinder is disabled

arjo129:
This adds a config to disable the RDI Pathfinder component which keeps causing our tests to timeout. Since we are not simulating this component I don't see why it should be enabled in our CI and tests.

### DIFF
--- a/docker/debug_integration/Dockerfile
+++ b/docker/debug_integration/Dockerfile
@@ -70,6 +70,9 @@ COPY . src/lrauv
 # Build image
 RUN [ "/bin/bash" , "-c" , "colcon build --cmake-args='-DBUILD_TESTING=true'" ]
 
+RUN [ "/bin/bash" , "-c" , "sed -i '48s/1/0/' ~/lrauv_ws/src/lrauv-application/Config/sim/Sensor.cfg"]
+RUN [ "/bin/bash" , "-c" , "sed -i '49s/1/0/' ~/lrauv_ws/src/lrauv-application/Config/sim/Sensor.cfg"]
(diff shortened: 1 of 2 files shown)


chapulina (commented):
I think this is harmless, but it was mentioned during the meeting that we should avoid disabling systems so the simulation is not very different from the hardware. So I think we should open an issue to revert this eventually.

I merged a PR upstream with Ben's fixes, which I believe target this. That's already in the Garden image. I'll publish a Fortress image with it to see if it fixes CI. Maybe we don't need this PR.

arjo129:
Cool I'll take a look at the new images then. Out of curiosity what was causing the early termination?

braanan:
Lack of position. The vehicle sources altitude from preloaded navigation charts — the Startup mission acquires GPS, which propagates to the vehicle navigator, etc. The Startup mission was omitted from the vehicle startup sequence because the surface is currently not supported. The fix I put in the MBARI repo partially restores some of the startup functionality (basically comms + GPS) but keeps the GoToSurface behavior that's part of that sequence commented out (until we have a surface boundary). 

One potential issue that I see is that if the vehicle is spawned at a depth > 1 m the radios will not retrieve GPS (normally on startup the vehicle drives to the surface and then activates the radios) — so in some cases the DVL may fail.

I'll take a look at adding the configured startup position from Config/workSite.cfg directly to the navigation charts, but these charts are also used in flight code so we might have to figure something else out.

arjo129:
Closing in favor of ustream change.

PR #117, opened 2021-12-14
Increase depth tolerance for spawn test

chapulina:
The test started failing, it looks like we're sinking more than usual.

### DIFF
--- a/lrauv_ignition_plugins/test/test_spawn.cc
+++ b/lrauv_ignition_plugins/test/test_spawn.cc
@@ -144,7 +144,7 @@ TEST(SpawnTest, Spawn)
   // and pitch are currently unstable, see
   // https://github.com/osrf/lrauv/issues/49
   double tightTol{1e-5};
-  double depthTol{1e-2};
+  double depthTol{0.1};
   double pitchTol{1e-2};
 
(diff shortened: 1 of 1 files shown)


arjo129 (approved):
Yeah I don't understand why though. We may need to investigate. I know in #89 that the vehicle is not sinking at all otherwise some other unit test would fail, but the spawned vehicle seems to sink... There may be yet another bug in the buoyancy plugin.

chapulina:
> Yeah I don't understand why though. We may need to investigate. 

Yeah I'm not sure either. I think the only thing that changed and could affect this is the upstream buoyancy fix, which is already released.

I'll merge this, but we still need to fix the expectations for `PitchMass`. If not with #112 , then some other way.

arjo129:
So I'm hoping that I can get #96 will go green by end of the day with the necessary fixes. That should allow pitchmass to work.

chapulina:
> I'm hoping that I can get #96 will go green by end of the day with the necessary fixes. That should allow pitchmass to work.

My concern is that #96 is pretty large and it builds on top of another PR. So it should take a while to get it in. I think it's worth it getting CI green before we do all those large changes.

PR #118, opened 2021-12-15
Addresses #62. Creates a GUI for tethys

### LINKED ISSUE #62: Tethys control GUI
The keyboard teleop is helpful, but a custom GUI widget for the Tethys would help getting an overview of the current status in one place.

Some quantities that would be nice to read / write:

* current values for each of the joints
* vehicle's pose
* buoyancy engine state
* drop weight state

arjo129:
Closes #62

TBH this is kind of ugly (I failed art in primary school), but this PR creates a GUI for all the control surfaces of the LRAUV.

tethys_gui

Feedback is welcome. Still trying to learn QML properly (thank god I'm not a front end dev).


chapulina (approved):
Thanks for this, it's so much easier to use than the keyboard teleop!

Future feature request: it would be good to update the values on the GUI from the state message, so that when we change the vehicle name, the GUI is updated to the current values for that vehicle.

> TBH this is kind of ugly

I figured you wouldn't mind some help with the layout, so I pushed directly to this branch on 6bc2a2f :wink: 

tethysgui

mabelzhang:
Just saw this while reviewing the Garden PR and hunted it down here. This is so useful!!!! Thanks for adding it!!

mabelzhang:
I think we should add something like this to the a `ign-gazebo` list of tutorials here https://ignitionrobotics.org/api/gazebo/6.3/tutorials.html
It'd need an example robot (maybe the model upstream is good enough?)

The control panel is a very good idea and great for general robotics entry-level learning of what a robot can do, by moving its joints. It could be part of the suite of beginner tutorials to Ignition (when we have those tutorial labels).

arjo129:
That sounds like a great idea. I'd be willing to do this.

mabelzhang:
@braanan Have you seen @arjo129's new plugin panel with sliders for the actuator controls? It's so nice!! It adds much more visual understanding and user interaction. I’d propose including it in your periodic report’s highlights.

I love this video game haha it's so fun.

My abstract art using rudder, elevator, thruster, VBS, and finally drop weight (didn't use mass shifter, oh well):

2021-12-17-032209_2400x2000_scrot_ControlPanelPlugin
2021-12-17-032227_2400x2000_scrot

Video of its creation (see 00:43 for sudden change in thruster, and 1:47 for drop weight):

https://user-images.githubusercontent.com/7608908/146514822-f04cac71-36af-430e-b25d-ee3e8a9ee2c1.mp4

braanan:
Very cool! Thanks for the hard work @arjo129!

braanan:
Just played around with this new feature - it's super nice! I think we can integrate it nicely with our operator training program to help pilots develop intuition for the how the vehicle responds to the actuators. One minor suggestion is to display the units in degrees and rpm instead of radians - those units might be more intuitive for end users.

mabelzhang:
We should get your UX team's feedback on the Ignition sim if we can steal
their time xD Julia was in our breakout room at the operator training.

PR #121, opened 2021-12-17
Update quit string on tests

chapulina:
That has changed on the new Docker images.

A separate "issue" is that the missions now don't terminate on critical failures, so they're running until the timeout - which causes CI to take longer and test expectations to fail (we get more drift, etc).


braanan (approved):

chapulina:
Cancelled CI after 1h 34 min. `PitchMass` didn't even have a chance to reach the mission timeout and the test timed out after 1500.23 sec.

chapulina:
Merging this, I'm working on a fix for the timeouts for a new PR.

Issue #4, opened 2021-08-20
Terminate lrauv-application cleanly


mabelzhang:
## Expected behavior

lrauv-application should terminate cleanly on `exit` or `quit` commands.

## Actual behavior

Currently, the lrauv-application does not terminate cleanly when running with Ignition. This only happens when running with Ignition.
When `quit` or `exit` is issued, the command prompt hangs:
```
> quit
>
```
Ctrl+C is needed to terminate the program.

Initial investigation found that ign-transport has a singleton node that is not meant to be terminated, and the destructors are clean.

There could be a mismatch in expectations, where ign-transport does not intend to terminate, but the lrauv-application's tear-down procedure expects all threads to terminate and waits for the ign-transport thread to merge, which never happens.

## Who even cares

Clean termination is useful for automated tests such as in osrf/lrauv_private#82, which are currently force-terminating the program.
It is also useful to know when all the log files are written and closed properly, which is guaranteed on a clean termination.

## Suggestions of how to start

@caguero and @braanan know better where exactly the lines are in the code in ign-transport and lrauv-application, respectively.

chapulina:
Here's the upstream issue: https://github.com/ignitionrobotics/ign-transport/issues/101

mabelzhang:
Wow, I think that issue was forgotten, because it never came up when we first looked into the issue.

chapulina:
I spent some time looking into it today. Not getting rid of the singleton completely, but tried to add a helper function to terminate all threads. After a couple hours I gave up :upside_down_face:

braanan:
Thanks for sharing and taking a look. I also tried creating the transport context on the heap and actively deleting in the destructor. That didn't work. The next thing I'd like to verify is if the issue happens when we only have one transport instance in the app - one thing we could try is creating a transport singleton in the application with accessor functions like we do with our LCM instance.

braanan:
This issue is resolved following the removal of `pthread_exit` in lrauv-application `main`.

Issue #39, opened 2021-10-05
Automated mission tests do not return clean command prompt


mabelzhang:
## Unexpected behavior
As discussed in https://github.com/osrf/lrauv/pull/82#issuecomment-910381371 , automated mission tests added in osrf/lrauv#82 would finish running, return the command prompt, but nothing typed at the prompt would be visible.
The typed content is there – if you hit enter, the typed stuff would get executed. It just doesn't show up.

User must type `reset` to get keyboard input to show up again.

## Measurement of success
The following tests return clean prompt on termination:
```
build/lrauv_ignition_plugins/test_mission_pitch_mass 
build/lrauv_ignition_plugins/test_mission_depth_vbs
build/lrauv_ignition_plugins/test_mission_yoyo_circle
```

hidmic:
This is no longer a problem. In any case, mission tests no longer live in this repository.

PR #122, opened 2021-12-17
Mission tests: kill LRAUV application after enough iterations

chapulina:
Thanks to @braanan , now the missions don't end up in critical errors anymore. Which means that they run until their timeouts, which are very large (hours).

Our tests had been relying on the missions ending relatively quickly until now, and with the latest fixes, they're running forever (I had to cancel GitHub actions that was running for 1.5 hours on #121).

Instead of updating the mission timeouts and relying on the LRAUV application terminating on its own, the idea on this PR is that the tests should know when they've received enough data and terminate the application then.

We've been killing the application anyway because of #4 , so now we're actually taking advantage of that.

### DIFF
--- a/lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
+++ b/lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
@@ -53,16 +53,18 @@ TEST_F(LrauvTestFixture, DepthVBS)
         lrauvRunning);
   });
 
-  int maxIterations{28000};
+  // Run enough iterations (chosen empirically) to reach steady state, then kill
+  // the controller
+  int targetIterations{28000};
   int maxSleep{100};
   int sleep{0};
-  for (; sleep < maxSleep && lrauvRunning && this->iterations < maxIterations; ++sleep)
+  for (; sleep < maxSleep && lrauvRunning && this->iterations < targetIterations; ++sleep)
   {
     igndbg << "Ran [" << this->iterations << "] iterations." << std::endl;
     std::this_thread::sleep_for(1s);
   }
   EXPECT_LT(sleep, maxSleep);
-  ASSERT_LT(maxIterations, this->tethysPoses.size());
+  ASSERT_LT(targetIterations, this->tethysPoses.size());
 
   LrauvTestFixture::KillLRAUV();
   lrauvThread.join();
@@ -70,7 +72,7 @@ TEST_F(LrauvTestFixture, DepthVBS)
   ignmsg << "Logged [" << this->tethysPoses.size() << "] poses" << std::endl;
 
   // Uncomment to get new expectations
-  // for (int i = 2000; i <= maxIterations; i += 2000)
+  // for (int i = 2000; i <= targetIterations; i += 2000)
   // {
   //   auto pose = this->tethysPoses[i];
   //   std::cout << "this->CheckRange(" << i << ", {"
--- a/lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
+++ b/lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
@@ -53,22 +53,21 @@ TEST_F(LrauvTestFixture, DepthVBS)
         lrauvRunning);
   });
 
+  int maxIterations{28000};
   int maxSleep{100};
   int sleep{0};
-  for (; sleep < maxSleep && lrauvRunning; ++sleep)
+  for (; sleep < maxSleep && lrauvRunning && this->iterations < maxIterations; ++sleep)
   {
     igndbg << "Ran [" << this->iterations << "] iterations." << std::endl;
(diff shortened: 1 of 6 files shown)


lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -53,23 +53,22 @@ TEST_F(LrauvTestFixture, DepthVBS)
         lrauvRunning);
   });
 
+  int maxIterations{28000};
```
mabelzhang, line 56: I realize this line is not new, but maybe a comment saying why specifically 28000, even if it's just empirically the number that works for all the missions? Same for other files.

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -53,23 +53,22 @@ TEST_F(LrauvTestFixture, DepthVBS)
...
   int sleep{0};
-  for (; sleep < maxSleep && lrauvRunning; ++sleep)
+  for (; sleep < maxSleep && lrauvRunning && this->iterations < maxIterations; ++sleep)
   {
     igndbg << "Ran [" << this->iterations << "] iterations." << std::endl;
     std::this_thread::sleep_for(1s);
   }
   EXPECT_LT(sleep, maxSleep);
-  EXPECT_FALSE(lrauvRunning);
+  ASSERT_LT(maxIterations, this->tethysPoses.size());
```
mabelzhang, line 65: I realize this line is not new either, but I want to check reader understanding. Is this assertion to make sure it has run for at least `maxIterations`? In that case, maybe it should be renamed to something like targetIterations, because at first I thought there's a typo, asserting a "max" is less than something. Same for other files.

mabelzhang (approved):
CI failing on `LrauvTestFixture.Command`?

The 4 mission tests in this PR pass for me locally. (Tested in Garden because I've deleted my Fortress workspace :sweat_smile:)

Locally, the terminal does not return cleanly for you too, right? I have to type `reset`, otherwise what I type doesn't show up. I feel like we've had this conversation before. Maybe I'll add it to #4 so I don't have to ask this again.

mabelzhang:
(I forgot to comment out the debug lines for the barycentric interpolation, so CI is now printing a lot of junk. Will open small PR)

mabelzhang:
Looks like `LrauvTestFixture.Command` is run twice, test 2 passed, and test 13 failed.

chapulina:
> Looks like LrauvTestFixture.Command is run twice, test 2 passed, and test 13 failed.

Ouch, two tests had the same name. Fixed name in e0aba90, and also tried to make the test more robust

chapulina:
> Locally, the terminal does not return cleanly for you too, right? I have to type reset, otherwise what I type doesn't show up. I feel like we've had this conversation before. Maybe I'll add it to #4 so I don't have to ask this again.

* Oh there's an issue just for that: https://github.com/osrf/lrauv/issues/39

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
chapulina: Documented on ad14dbf

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
chapulina: True, went with `targetIterations`:  ad14dbf

mabelzhang (approved):
CI passing, looks good.

:sweat_smile: Ha okay so we did have that conversation before.

braanan:
Thanks @chapulina! I think this will work for now, but in general, terminating the app while it's running (without properly uninitializing) might cause some issues. 

Particularly, on the next startup the vehicle will throw a critical failure stating that the last restart was unintentional - this mechanism was put out in place to notify the operator in case the application crashes while the vehicle is deployed at sea (the watchdog timer automatically reboots the system if the app crashes and the app is launched at startup). After an unintentional boot the app will ignore any persisted configuration settings that are found in `Data/persisted.cfg` - this happens so the operator gets an opportunity to reconfigure the vehicle if a config that was added at run time is causing issues. 

The way this works is that the app creates an empty file in the Data folder called `running`, which is removed as part of the normal teardown. That way if `Data/running` is present at startup the app can deduce that an unintentional reboot happened. 

I don't think this matters for CI testing, but wanted to make you aware of this. Once we resolve #4 we should adjust mission timeouts to work with CI expectations.

Thanks!

chapulina:
Thanks for all the context, @braanan . For automated tests, I think ultimately it would be ideal if we could start the controller via code, as opposed to launching the LRAUV CLI. This way we wouldn't need to fork a process, which brings complications like #39. I don't know how feasible or desirable it would be to expose a public API for the controller, but I thought I should mention it anyway.

PR #126, opened 2021-12-22
Remove miniature scale from science data and fix segfault

chapulina:
* Broken off #88 

Tests were crashing for that PR, while investigating, I noticed it happened because the `MINIATURE_SCALE` had been removed.

This PR removes the miniature scale and prints a warning instead of segfaulting, but I haven't looked closely into the barycentric interpolation to see why the miniature scale caused the issue, could you take a look @mabelzhang ?

I also added other changes from #88 which help us not need the miniature scale:

* The invisible ground plane makes it possible to easily scroll to a far away distance
* This can be seen with the 300kmx300km grid that is also added here
* The increased camera clipping distances are also needed to see the entire grid from a distance
* Added the camera controls plugin so the user can easily go back to the initial camera angle

miniature_scale

### DIFF
--- a/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
+++ b/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
@@ -104,9 +104,9 @@ class tethys::ScienceSensorsSystemPrivate
     const std::vector<int> &_inds,
     std::vector<float> &_new);
 
-  /// \brief Sort vector in-place, keeping track of indices
+  /// \brief Sort vector, store indices to original vector after sorting.
+  /// Original vector unchanged.
   /// \param[in] _v Vector to sort
-  /// \param[out] _v Sorted vector
   /// \param[out] _idx Indices of original vector in sorted vector
   public: template<typename T>
   void SortIndices(
@@ -880,7 +880,7 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
     return std::numeric_limits<float>::quiet_NaN();
   }
 
-  // Sort points
+  // Sort points and store the indices
   std::vector<float> xsSorted;
   std::vector<size_t> xsSortedInds;
   for (int i = 0; i < _xs.size(); ++i)
@@ -888,6 +888,11 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
     xsSorted.push_back(_xs(i));
   }
   SortIndices(xsSorted, xsSortedInds);
+  // Access sorted indices to get the new sorted array
+  for (int i = 0; i < xsSortedInds.size(); ++i)
+  {
+    xsSorted[i] = _xs(xsSortedInds[i]);
+  }
 
   int ltPSortedIdx{-1};
   int gtPSortedIdx{-1};
@@ -909,13 +914,23 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
       break;
     }
   }
+
+  // Sanity check
   if (ltPSortedIdx < 0 || ltPSortedIdx >= xsSortedInds.size() ||
       gtPSortedIdx < 0 || gtPSortedIdx >= xsSortedInds.size())
   {
-    ignwarn << "Failed to find consecutive elements in sorted vector. Indices ["
-            << ltPSortedIdx << "] / [" << gtPSortedIdx
-            << "] for vector sized [" << xsSortedInds.size()
-            << "]. Cannot interpolate." << std::endl;
+    ignwarn << "1D linear interpolation: cannot find pair of consecutive "
+      << "neighbors that query point lies between. Cannot interpolate. "
+      << "(This should not happen!)"
+      << std::endl;
+    if (this->DEBUG_INTERPOLATE)
+    {
+      igndbg << "Neighbors: " << std::endl << _xs << std::endl;
+      igndbg << "Sorted neighbors: " << std::endl;
+      for (int i = 0; i < xsSorted.size(); ++i)
+        igndbg << xsSorted[i] << std::endl;
+      igndbg << "Query point: " << std::endl << _p << std::endl;
+    }
     return std::numeric_limits<float>::quiet_NaN();
   }
 
@@ -927,9 +942,13 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
   int ltPIdx = xsSortedInds[ltPSortedIdx];
   int gtPIdx = xsSortedInds[gtPSortedIdx];
 
+  // Sanity check
   if (ltPIdx >= _values.size() || gtPIdx >= _values.size())
   {
-    ignwarn << "Bad indices. Cannot interpolate." << std::endl;
+    ignwarn << "1D linear interpolation: mapping from sorted index to "
+      << "original index resulted in invalid index. Cannot interpolate. "
+      << "(This should not happen!)"
+      << std::endl;
     return std::numeric_limits<float>::quiet_NaN();
   }
   float result = ltPWeight * _values[ltPIdx] + gtPWeight * _values[gtPIdx];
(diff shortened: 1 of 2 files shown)


lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -925,6 +909,15 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
       break;
     }
   }
+  if (ltPSortedIdx < 0 || ltPSortedIdx >= xsSortedInds.size() ||
```
mabelzhang, line 912: Is this happening locally for you? The sanity check above that prints
```
    ignwarn << "1D linear interpolation: query point lies on one side of all "
      "interpolation points. Cannot interpolate." << std::endl;
```
is supposed to preclude this from happening. If p is not on one side of all the points, then it has to be in between two points. If this is happening, then a case isn't being checked, and it should be checked before this for-loop, instead of checking afterwards.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -933,6 +926,12 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
   // Retrieve indices of sorted elements in original array
   int ltPIdx = xsSortedInds[ltPSortedIdx];
   int gtPIdx = xsSortedInds[gtPSortedIdx];
+
+  if (ltPIdx >= _values.size() || gtPIdx >= _values.size())
```
mabelzhang, line 930: This too should not happen. I'd like to reproduce to see why it happens and check it earlier in the code.

mabelzhang (commented):
Thank you for debugging! Are you able to find data to test it locally? If I just delete the added if-statements, it still doesn't seg fault or hang for me.

I checked the hanging test at https://github.com/osrf/lrauv/pull/88/checks?sha=fa28e234918d0c2a86232d272e931095e2c1600c , and the output 
```
10: [Wrn] [ScienceSensorsSystem.cc:858] 1D linear interpolation: query point lies on one side of all interpolation points. Cannot interpolate.
```
suggests that it returned NaN and something continued looping, possibly one of the loops in PostUpdate():
```
  for (auto &[entity, sensor] : this->entitySensorMap)
```

On a side note, the bad news of looking at this actual data is that, I now remember why we needed the per-slice searches as opposed to barycentric. Ughhhhhh. It's the z resolution being significantly smaller than x and y! The 4 NNs are all in 1 line along z. We might have to go back to trilinear interpolation. Ugh. At least I haven't deleted those branches.

To make our lives easier, we might have to take a step back and make the strict assumption that input indices are sorted, like in netCDF.

On a related note, debugging the interpolation will be significantly easier if we can publish a visual Marker from the ScienceSensorsSystem plugin. Anything against that?

The last thing I left off in the trilinear interpolation branch is that the found 4 NNs didn't satisfy the assumption that they need to be in a rectangular prism, which was the reason I moved to the less restrictive barycentric interpolation. I would publish a Marker to see if there are any bugs to shake out, before giving up and rewinding to assume all the input indices are sorted.

Verdict: real data is ugly.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
chapulina: Yeah many tests are failing because of this, both locally for me and on CI. You can see how the "Failed to find consecutive elements in sorted vector." message is printed on CI multiple times. You can test with `test_controller` for example.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
chapulina: This one indeed doesn't happen for me, but while debugging I figured I'd add a check just in case. It may fail in the future...

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
mabelzhang: Hmm I still cannot reproduce a crash. I commented out this block and the block below (diff below), and I ran almost all the `test_*` executables locally, both the standalone ones and the `test_mission_*` ones. I thought maybe you meant all the mission tests fail, but those pass locally for me too. Most of them just print the blue messages. Yoyo prints the one I saw on old CI:
```
[Wrn] [ScienceSensorsSystem.cc:878] 1D linear interpolation: query point lies on one side of all interpolation points. Cannot interpolate.
```
but it passes too.

My only diff is this:
```diff
$ git diff chapulina/barycentric_segfault
diff --git a/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc b/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
index a3cfff4..f222125 100644
--- a/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
+++ b/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
@@ -196,7 +196,7 @@ class tethys::ScienceSensorsSystemPrivate
 
   /// \brief Debug printouts for interpolation. Will keep around at least until
   /// interpolation is stable.
-  public: const bool DEBUG_INTERPOLATE = false;
+  public: const bool DEBUG_INTERPOLATE = true;
 
   ///////////////////////////////
   // Variables for coordinate system
@@ -909,6 +909,7 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
       break;
     }
   }
+  /*
   if (ltPSortedIdx < 0 || ltPSortedIdx >= xsSortedInds.size() ||
       gtPSortedIdx < 0 || gtPSortedIdx >= xsSortedInds.size())
   {
@@ -918,6 +919,7 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
             << "]. Cannot interpolate." << std::endl;
     return std::numeric_limits<float>::quiet_NaN();
   }
+  */
 
   // Normalize the distances to ratios between 0 and 1, to use as weights
   float ltPWeight = ltPDist / (gtPDist + ltPDist);
@@ -927,11 +929,13 @@ float ScienceSensorsSystemPrivate::BarycentricInterpolate(
   int ltPIdx = xsSortedInds[ltPSortedIdx];
   int gtPIdx = xsSortedInds[gtPSortedIdx];
 
+  /*
   if (ltPIdx >= _values.size() || gtPIdx >= _values.size())
   {
     ignwarn << "Bad indices. Cannot interpolate." << std::endl;
     return std::numeric_limits<float>::quiet_NaN();
   }
+  */
   float result = ltPWeight * _values[ltPIdx] + gtPWeight * _values[gtPIdx];
 
   if (this->DEBUG_INTERPOLATE)
```

mabelzhang: Oh, `test_controller` actually does seg fault. I guess I didn't run the right one. Most of the others passed. I'll look into what case `test_controller` activates.

mabelzhang: It was a bug, fixed in b8641f02224ef2b9172301234277e73871919dbe
Turns out I never actually stored the sorted array, only stored the indices :sweat_smile: 
Now you should not see the bumpers printed. I changed the messages to be more precise what's happening, and added that they aren't supposed to happen so the user (future me?) will hopefully look for bugs.

mabelzhang (approved):
Should be fixed in b8641f0
So the cause was a pretty important bug - I never stored the array after sorting, only stored the sorted indices, heh.
I'm gonna approve my own fix :sweat_smile: Try it and see if the new warnings still print for you - they shouldn't anymore.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
chapulina: Nice! Thanks!

chapulina:
Warnings are gone! Thanks for fixing the bug :D

PR #142, opened 2022-01-11
A suggestion for a simpler search algorithm for the interpolation

tfoote:
Here's a suggestion for accelerating/simplifying the search algorithm. It's mostly focused on reducing the complexity of the search. There's now just two slice ranges, followed by a closest neighbor search in each of those slices. 

I removed unused parameters: c9130c5dacc69a50918ba3b2dbcf062c5ac9c8f6

The biggest thing that I don't have an immediate answer to is how to set what is currently a magic number of the width of the slice. My first vale of 1m didn't find anything below. And my second try at 100m finds way to many. We should find a way to query this resolution from the dataset or parameterize it somehow. 

Looking at the output it appears that 5m resolution is our current resolution. So I've set it to that.

### DIFF
--- a/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
+++ b/lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
@@ -89,15 +89,11 @@ class tethys::ScienceSensorsSystemPrivate
   /// \brief Find XYZ locations of points in the two closest z slices to
   /// interpolate among.
   /// \param[in] _pt Location in space to interpolate for
-  /// \param[in] _inds Indices of nearest neighbors to _pt
-  /// \param[in] _sqrDists Distances of nearest neighbors to _pt
   /// \param[out] _interpolators1 XYZ points on a z slice to interpolate among
   /// \param[out] _interpolators2 XYZ points on a second z slice to interpolate
   /// among
   public: void FindTrilinearInterpolators(
     pcl::PointXYZ &_pt,
-    std::vector<int> &_inds,
-    std::vector<float> &_sqrDists,
     std::vector<pcl::PointXYZ> &_interpolators1,
     std::vector<pcl::PointXYZ> &_interpolators2);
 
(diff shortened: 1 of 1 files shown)


tfoote:
The z slice at 5m worked:

```
1: [Dbg] [ScienceSensorsSystem.cc:757] 295536 points in full cloud
1: [Dbg] [ScienceSensorsSystem.cc:763] 1st z slice 18471 points
1: [Dbg] [ScienceSensorsSystem.cc:841] Found 4 neighbors.
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (1.39148e-08, -1.1001e-09, -0), distance 1.07004e-06 m
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (-1812.03, 0.184071, -0), distance 1812.03 m
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (1812.03, 0.184071, -0), distance 1812.03 m
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (1.37634e-08, -2218.66, -0), distance 2218.66 m
1: [Dbg] [ScienceSensorsSystem.cc:775] 2nd z slice 18471 points
1: [Dbg] [ScienceSensorsSystem.cc:841] Found 4 neighbors.
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (1.39148e-08, -1.1001e-09, -5), distance 5 m
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (-1812.03, 0.184071, -5), distance 1812.04 m
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (1812.03, 0.184071, -5), distance 1812.04 m
1: [Dbg] [ScienceSensorsSystem.cc:849] Neighbor at (1.37634e-08, -2218.66, -5), distance 2218.66 m
1: [Dbg] [ScienceSensorsSystem.cc:884] diff: 
```

It's hard to find things in the logs now that they're truncated at 55thousand lines

The CI looks like YoYoCircle failed. I'm guessing this hasn't been rebased recently so is still flakey.

mabelzhang:
So, for speedy reviews, I would really like to have new changes be based on #127, instead of #83, because #83 is not meant to work without #127, so it's very hard to judge whether things are correct based on #83. It has a FIXME that's a big problem, which means the results returned in #83 are actually incorrect - the indices of the points aren't mapped back the original point cloud. So everything downstream from `FindTrilinearInterpolators` in #83 are just undefined behavior.

I would prefer to address the comments in #83 individually, merge that, and then move additional suggestions to #127, where we have actual interpolation that we trust much more, and have more data to test with.

There had been so many failing edge cases that we had to address - which #127 addresses, I'm very hesitant to look at one data output and declare things working.

tfoote:
Replaced this with #156

PR #143, opened 2022-01-12
Spawn vehicle at requested orientation

chapulina:
Part of 

* https://github.com/osrf/lrauv/issues/48

Requires

* https://github.com/ignitionrobotics/ign-gazebo/pull/1263
* https://github.com/ignitionrobotics/ign-msgs/pull/212
* Forward-port: https://github.com/ignitionrobotics/ign-gazebo/pull/1293

This is a subset of

* #81 

---

Make use of the roll, pitch and heading fields on the init message when spawning a vehicle.

I also changed the test to run paused, so we don't have the physics affecting the spawn pose and we can check it more precisely.

This takes an alternative approach to #81. Instead of rotating the model to face North by default, we introduce offsets when receiving spawn commands. This means we must be mindful of the fact that the vehicle faces West by default all the time when looking at poses, which is extremely error prone. I'd much rather change the model to match NED / FSK by default, but as explained in https://github.com/osrf/lrauv/issues/48#issuecomment-1018063778, there are physics issues with using a Z-down model that we'll need more time to track down.


chapulina:
I forgot that the orientation is in NED, so this can't escape #81 .

chapulina:
Upstream's forward-port was merged. After tonight's nightlies are built, we should retrigger CI here.

chapulina:
Ahh last night's `ign-gazebo7` nightly failed:

https://build.osrfoundation.org/job/ign-gazebo7-debbuilder/159/

chapulina:
Yay, the latest nightlies are out and CI is :green_circle:

arjo129 (approved):
Thanks for these docs.

PR #154, opened 2022-01-27
tethys_equipped face North by default

chapulina:
* Part of #48
* Builds on top of #143 

This PR updates the `tethys_equipped` model so that it faces North by default, i.e. zero orientation means it faces North, to match what the controller expects. Before this PR, the model was facing West by default.

Before this PR | After this PR
-- | --
image | image

Summary:

* :warning:  `tethys_equipped` faces North / +Y, but `tethys` by itself still faces -X / West
* The state message has been updated to publish pos and RPH in NED (see tests)
* :warning: The model's base link (canonical link) still faces the model's -X, which is the world's -Y at pose zero


lrauv_ignition_plugins/src/TethysCommPlugin.cc
```diff
@@ -173,6 +173,13 @@ void ROSToFSK(ignition::math::Vector3d &_vec)
   _vec.Z(-_vec.Z());
 }
 
+// Convert a pose in ENU to NED.
+ignition::math::Pose3d ENUToNED(ignition::math::Pose3d &_enu)
```
arjo129, line 177: Minor knit: document argument. Argument can also be marked `const`.

lrauv_ignition_plugins/test/test_mission_pitch_and_depth_mass_vbs.cc
```diff
@@ -83,14 +93,14 @@ TEST_F(LrauvTestFixture, PitchDepthVBS)
     EXPECT_GT(pose.Pos().Z(), -21.5);
 
     // Vehicle should exhibit minimal lateral translation.
-    EXPECT_NEAR(pose.Pos().X(), 0, 10); // FIXME(arjo): IMPORTANT!!
-    EXPECT_NEAR(pose.Pos().Y(), 0, 1e-1);
+    EXPECT_NEAR(pose.Pos().X(), 0, 1e-3);
```
arjo129, line 96: This tolerance may be a bit too tight. I think we can expect some amount of drift when running these missions especially since there is no control of the thruster.

arjo129 (approved):
Thanks for getting this in! I like the suffixing of coordinate systems to variable names, gonna start doing that. This looks good to me.

lrauv_ignition_plugins/src/TethysCommPlugin.cc
chapulina: Good call 92dcef0894521140d33c562458ea86c710173b13

lrauv_ignition_plugins/test/test_mission_pitch_and_depth_mass_vbs.cc
chapulina: I was able to reduce it to `1e-5` in 92dcef0894521140d33c562458ea86c710173b13. Note that his corresponds to the previous Y, which had `1e-1`. I think sub-millimeter accuracy is pretty good here :100:

chapulina:
Thanks for the review, @arjo129 . I've addressed all of your comments. This PR builds on top of #143 because without that the spawn test would fail. I think it's best to merge that one first so that it's easier to review.

lrauv_ignition_plugins/test/test_mission_pitch_and_depth_mass_vbs.cc
chapulina: I had to increase it to`2e-3` in c752c42a08082347b00157a11559b7efa57c256a because it was failing on CI. That's still less than the `1e-1` that we had before, but not as nice as `1e-5`...

PR #155, opened 2022-01-27
Update posDot state velocity to NED

chapulina:
* Part of #48
* Builds on top of #154

Summary

* Update message documentation for `speed_`  and `posDot_`
* Fix their values
* Add tests


arjo129 (approved):
Thanks for these cleanups.

PR #141, opened 2022-01-08
use a dist-upgrade not just upgrade

tfoote:
The upgrade will skip thing that require new packages. Such as new ignition libraries which are what we're looking for going to the nightlies.

This is a follow up to https://github.com/osrf/lrauv/pull/128#issuecomment-1006103394 https://github.com/osrf/lrauv/commit/e1ebc3a7e3b4f2e4664a781f05be823ebce46e6a

Reproducing the upgrade it skips a lot
<details>

```
root@6e59e94e5c93:/home/developer/lrauv_ws/src/lrauv-application# apt-get upgrade -y
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Calculating upgrade... Done
The following packages have been kept back:
  ignition-garden libegl-mesa0 libegl-mesa0:i386 libgbm1 libgbm1:i386 libgl1-mesa-dri libgl1-mesa-dri:i386 libglapi-mesa
  libglapi-mesa:i386 libglx-mesa0 libglx-mesa0:i386 libignition-fuel-tools8 libignition-fuel-tools8-dev libignition-gazebo7
  libignition-gazebo7-dev libignition-gazebo7-plugins libignition-gui7 libignition-gui7-dev libignition-launch6 libignition-launch6-dev
  libignition-msgs9 libignition-msgs9-dev libignition-rendering7 libignition-rendering7-core-dev libignition-rendering7-dev
  libignition-rendering7-ogre1 libignition-rendering7-ogre1-dev libignition-rendering7-ogre2 libignition-rendering7-ogre2-dev
  libignition-sensors7 libignition-sensors7-air-pressure libignition-sensors7-air-pressure-dev libignition-sensors7-altimeter
  libignition-sensors7-altimeter-dev libignition-sensors7-camera libignition-sensors7-camera-dev libignition-sensors7-core-dev
  libignition-sensors7-depth-camera libignition-sensors7-depth-camera-dev libignition-sensors7-dev libignition-sensors7-force-torque
  libignition-sensors7-force-torque-dev libignition-sensors7-gpu-lidar libignition-sensors7-gpu-lidar-dev libignition-sensors7-imu
  libignition-sensors7-imu-dev libignition-sensors7-lidar libignition-sensors7-lidar-dev libignition-sensors7-logical-camera
  libignition-sensors7-logical-camera-dev libignition-sensors7-magnetometer libignition-sensors7-magnetometer-dev
[...]

PR #157, opened 2022-02-08
Minor fixes to empty_world/Dockerfile

hidmic:
Precisely what the title says. A very minor issue with newline escaping in our Dockerfile.

### DIFF
--- a/docker/empty_world/Dockerfile
+++ b/docker/empty_world/Dockerfile
@@ -67,6 +67,7 @@ RUN apt-get -qq update && apt-get -q -y install \
   wget \
   python3-vcstool \
   python3-colcon-common-extensions \
+  && apt-get dist-upgrade -y \
   && rm -rf /var/lib/apt/lists/* \
   && apt-get -qq clean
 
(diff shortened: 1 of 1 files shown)


hidmic:
Hmm integration tests are failing hard on `master`.

arjo129:
Something must have broken upstream *all* our tests are segfaulting.

arjo129 (approved):
This does not effect our core work flow or CI and seems to work for me.

tfoote (dismissed):
This looks correct. The change likely has to do with triggering a rebuild and possibly a partial debian package update. 

We should probably be doing a dist-upgrade when we install from the osrfoundation.org repos. https://github.com/osrf/lrauv/pull/141 I think that upstream should be pinning dependencies or declaring conflicts to force a full upgrade when you enable the osrfoundation repos to avoid incompatibility.

tfoote (commented):
The dist-upgrades look appropriate. If this is still failing I think that we have a bigger problem than the potential conflicting binary packages.

arjo129:
I think the changes have been applied. Lets merge this as I realize the first instruction in our README is broken.

arjo129 (approved):

PR #158, opened 2022-02-08
Add AHRS to Tethys equipped model

hidmic:
Combining upstream IMU and magnetometer sensors. This is lacking a test but perhaps this is best tested as part of #151 e.g. make the AUV describe a known trajectory and sample IMU and magnetometer measurements along that trajectory to ensure values make sense.

### DIFF
--- a/lrauv_ignition_plugins/test/test_ahrs.cc
+++ b/lrauv_ignition_plugins/test/test_ahrs.cc
@@ -65,10 +65,10 @@ class AhrsTest : public ::testing::Test
                                     "worlds", "buoyant_tethys.sdf"));
 
     this->node.Subscribe(
-        "/model/tethys/ahrs/imu", &AhrsTest::OnImuMessage, this);
+        "/tethys/ahrs/imu", &AhrsTest::OnImuMessage, this);
 
     this->node.Subscribe(
-        "/model/tethys/ahrs/magnetometer",
+        "/tethys/ahrs/magnetometer",
         &AhrsTest::OnMagnetometerMessage, this);
 
     this->fixture->OnConfigure(
@@ -194,9 +194,9 @@ class AhrsTest : public ::testing::Test
 // Acceleration due to gravity
 static constexpr double g0{9.8};  // m/s^2
 // Earth magnetic field
-static constexpr double Be{6.0e-6};  // T
-static constexpr double Bn{2.3e-5};  // T
-static constexpr double Bu{-4.2e-5};  // T
+static constexpr double Be{5.5645e-6};    // T
+static constexpr double Bn{22.8758e-6};   // T
+static constexpr double Bu{-42.3884e-6};  // T
 
 static constexpr double angularVelocityTolerance{1e-4};
 static constexpr double linearAccelerationTolerance{1e-4};
--- a/lrauv_ignition_plugins/test/test_ahrs.cc
+++ b/lrauv_ignition_plugins/test/test_ahrs.cc
@@ -1,0 +1,527 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+/*
+ * Development of this module has been funded by the Monterey Bay Aquarium
+ * Research Institute (MBARI) and the David and Lucile Packard Foundation
+ */
+
+#include <gtest/gtest.h>
+
+#include <ignition/common/Console.hh>
+#include <ignition/common/Filesystem.hh>
+#include <ignition/gazebo/components/AngularVelocityCmd.hh>
+#include <ignition/gazebo/components/LinearVelocityCmd.hh>
+#include <ignition/gazebo/components/Physics.hh>
+#include <ignition/gazebo/components/Pose.hh>
+#include <ignition/gazebo/components/PoseCmd.hh>
+#include <ignition/gazebo/components/Recreate.hh>
+#include <ignition/gazebo/components/Static.hh>
+#include <ignition/gazebo/Model.hh>
+#include <ignition/gazebo/TestFixture.hh>
+#include <ignition/gazebo/Types.hh>
+#include <ignition/gazebo/Util.hh>
+#include <ignition/gazebo/World.hh>
+#include <ignition/math/Pose3.hh>
+#include <ignition/math/Quaternion.hh>
+#include <ignition/math/Vector3.hh>
+#include <ignition/msgs/imu.pb.h>
+#include <ignition/msgs/magnetometer.pb.h>
+#include <ignition/msgs/Utility.hh>
+#include <ignition/transport/Node.hh>
+
+#include <chrono>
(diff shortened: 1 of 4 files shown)


hidmic:
@braanan do we have a reference AHRS system from which to derive a noise model? Not that we have a lot of options -- we have gaussian or gaussian.

Also, I'd really need access to all of our repositories to complete AHRS integration.

braanan:
I think we'll go with Gaussian then. Here's a link to the Sparton AHRS-M2 we have installed across the fleet: https://www.spartonnavex.com/product/ahrs-m2/ All the info you need should be in the datasheet.

As for access, all checks are cleared — just waiting for final approval from our MT.

hidmic:
Alright, after much digging, I think I extracted all that can be extracted from the datasheet. There's no characterization of the magnetometer whatsoever, and IIUC I'm lacking data to use the bias drift model in Ignition.

lrauv_ignition_plugins/test/test_ahrs.cc
```diff
@@ -0,0 +1,527 @@
...
+    const ignition::math::Quaterniond &_qa,
+    const ignition::math::Quaterniond &_qb,
+    const double _tolerance)
+{
+  // Use geodesic distance between quaternions
+  return std::acos(2. * std::pow(_qa.Dot(_qb), 2) - 1.) < _tolerance;
+}
+
+//////////////////////////////////////////////////
+TEST_F(AhrsTest, FrameConventionsAreCorrect)
```
hidmic, line 216: @braanan @arjo129 at first I tried to break this test into smaller ones, but I'm unable to instantiate `ignition::gazebo::TestFixture` more than once in the same process. It aborts with `CHECK failed: GeneratedDatabase()->Add(...)`, which appears to be coming from the shared library that bears `lrauv`
protobuf messages. I suspect the Ignition plugin system is attempting to reload it upon fixture construction (see [here](https://groups.google.com/g/protobuf/c/RZRCnOywdSk/m/s5QvifLCBgAJ?utm_medium=email&utm_source=footer)).

lrauv_ignition_plugins/test/test_ahrs.cc
```diff
@@ -0,0 +1,527 @@
...
+          const std::shared_ptr<const sdf::Element> &,
+          ignition::gazebo::EntityComponentManager &_ecm,
+          ignition::gazebo::EventManager &)
+      {
+        ignition::gazebo::World world(
+            ignition::gazebo::worldEntity(_ecm));
+        ignition::gazebo::Entity tethysEntity =
+            world.ModelByName(_ecm, "tethys");
+        using ignition::gazebo::components::Static;
+        _ecm.CreateComponent(tethysEntity, Static(true));
```
hidmic, line 85: @braanan @arjo129 I forced the model to be static to avoid motion whenever I put the AUV in unusual positions. Unfortunately, it also means we are not testing AHRS output when moving. It'd be tricky nonetheless. Hydrodynamics modelling gets in the way and it's hard to tell what the angular velocity and linear acceleration is going to be.

lrauv_ignition_plugins/test/test_ahrs.cc
arjo129: Oh yeah I forgot to mention I think we keep each test isolated in a separate process as there was a bug in the way we are linking the messages. For whatever reason if we rely on external messages google protobuf will go :boom:. I have not been able to isolate exactly why this is the case.

lrauv_ignition_plugins/test/test_ahrs.cc
arjo129: I think we agreed that as long as the AHRS is reporting the correct coordinate frame we should hope for the best with the gyro.

hidmic: That's right. Thing is, I cannot test for gyro frame conventions if it's output is always zero. I guess we _are_ testing that the body frame is correct for the accelerometer so it should be good for the gyro too.

arjo129 (approved):
Hey @hidmic thanks for the tests and the imu. I think this is good to merge. For integration one possibility is to use the [`lrauv_state`](https://github.com/osrf/lrauv/blob/main/lrauv_ignition_plugins/proto/lrauv_state.proto) to pass the imu outputs. This structure is populated in [`TethysCommPlugin.cc`](https://github.com/osrf/lrauv/blob/main/lrauv_ignition_plugins/src/TethysCommPlugin.cc). In particular the `ratePQR` field may be of interest.

arjo129:
@hidmic Should I merge this?

hidmic:
> @hidmic Should I merge this?

@arjo129 need to make sure I can integrate it with the other half of the codebase first. Hopefully this week.

hidmic:
@braanan @arjo129 this one's ready for a final review. Tests are passing locally. Integration with the rest of the code seems functional (though I see Tethys going for a run as #167 reports).

hidmic:
Integration tests are failing upon installing `sdformat` :thinking:

braanan:
@hidmic looks like there's a reverse sign in the accelerometer data - FSK should have positive down. 

I'm seeing:
```
2022-03-16T18:57:25.334Z,1647457045.334 [AHRS_M2](INFO): Accel:   X: 3.318425, Y: 0.055010, Z: -9.219960 (m/s2)
```

hidmic:
@braanan my (perhaps vague) understanding of MEMS IMUs is that when standing still under constant gravity, its proof mass is displaced as if it was accelerating in the opposite direction (ie. upwards). This matches Ignition's IMU sensor behavior (see [here](https://github.com/ignitionrobotics/ign-sensors/blob/f2eb097977c35cd9ac6b81d5a5b51dd5bfeb94e7/src/ImuSensor.cc#L231) and gravity defaults [here](https://github.com/ignitionrobotics/sdformat/blob/c496c5bb787c25c6f3c2f5d10304da2a9efaafd6/src/World.cc#L57)) 

Having said that, I have no first-hand experience with Sparton's AHRS-M2 so perhaps it behaves differently.

braanan:
It does. Here's the output from an actual vehicle
```
2022-03-16T19:08:03.896Z,1647457683.896 [AHRS_M2](INFO): PITCH: -0.384570, ROLL: -3.744142, YAW: 213.573527 (deg)
2022-03-16T19:08:03.897Z,1647457683.897 [AHRS_M2](INFO): Accel:   X: 0.064123, Y: -0.640986, Z: 9.850662 (m/s2)
```
I'm not sure if this output format is true to all AHRS models or unique to the M2. I'll check, since that will inform where we should implement the sign flip. If it's indeed true for all AHRS models we can see about reversing the sign in the plugin, otherwise, the sign flip should go in the AHRS_M2 device driver in the lrauv-application.

braanan:
OK, I've asked around and it seems like there's not much consistency between vendors, so I'll add the sign flip in the AHRS_M2 driver with a comment.

hidmic:
> OK, I've asked around and it seems like there's not much consistency between vendors, so I'll add the sign flip in the AHRS_M2 driver with a comment.

@braanan hmm, will that work? It's not that the linear acceleration Z-axis is flipped, only gravity is. I'll go ahead and guess that the real AHRS is estimating gravity's magnitude and direction and re-applying it to provide these acceleration measurements. Maybe we have to do the same in the plugin :/

braanan:
@hidmic Looks like the imu plugin assumes that the linear accelarations (and the sensor) are alligned with the ENU frame of refrence. As far as I can tell, the rotation to NED is only applied to the oriantation data [here](https://github.com/ignitionrobotics/ign-sensors/blob/f2eb097977c35cd9ac6b81d5a5b51dd5bfeb94e7/src/ImuSensor.cc#L262), so we'll have to add that rotation to NED (FSK) to the linear accelarations and angular velocities in the plugin or take care of that on the LRAUV side. Maybe I'm missing something. 

Since this is an upstream plugin the second option will be simpler I think.

hidmic:
@braanan may I get an approval?

braanan (approved):

PR #160, opened 2022-02-28
Refactor Test Fixture to support different worlds 

arjo129:
Currently, the `LrauvTestFixture` allows us to only load `BuoyantTethys.sdf`. This PR refactors the test fixture. It creates two test classes:
* `LrauvTestFixtureBase` - An abstract base class which contains all the common functionality.
* `LrauvTestFixture` - Inherits from `LrauvTestFixtureBase` and loads the `buoyant_tethys.sdf`.

### DIFF
--- a/lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
+++ b/lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
@@ -50,9 +50,8 @@ double angleDiff(double _a, double _b)
   return diff.Radian();
 }
 
-/// \brief Convenient fixture that provides boilerplate code common to most
-/// LRAUV tests. This is an abstract class that does not provide any concrete
-/// worlds.
+/// \brief Abstract base class for fixture that provides boilerplate code common
+/// to most LRAUV tests.
 class LrauvTestFixtureBase : public ::testing::Test
 {
   // Setup the specified world.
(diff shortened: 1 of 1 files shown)


lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
```diff
@@ -50,12 +50,13 @@ double angleDiff(double _a, double _b)
...
-/// LRAUV tests.
-class LrauvTestFixture : public ::testing::Test
+/// \brief Abstract base class for fixture that provides boilerplate code common
+/// to most LRAUV tests.
+class LrauvTestFixtureBase : public ::testing::Test
 {
-  // Documentation inherited
-  protected: void SetUp() override
+  // Setup the specified world.
+  // \param[in] _worldName Name of the world to load.
```
chapulina, line 58:
```suggestion
  /// \param[in] _worldName Name of the world to load.
```

lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh
```diff
@@ -302,4 +303,15 @@ class LrauvTestFixture : public ::testing::Test
   /// \brief Publishes commands
   public: ignition::transport::Node::Publisher commandPub;
 };
+
+
+/// \brief Loads the default "buyant_tethys.sdf" world.
+class LrauvTestFixture : public LrauvTestFixtureBase
+{
+  /// Documentation inherited
```
chapulina, line 311:
```suggestion
  // Documentation inherited
```

So, within the Ignition codebase, we leave "Documentation inherited" comments to explain that we're not adding documentation on purpose. When it's not added, the documentation is taken from the base class. But if the comment is added with three `///`, then that actually overrides the base class' doc block and writes "Documentation inherited" as the documentation.

Not important here since we're not generating docs for these classes, but I thought I should mention

chapulina (approved):

PR #161, opened 2022-02-28
Add surface to buoyant_tethys

arjo129:
Depends on #160

This PR adds a surface at z=0 to the default `buoyant_tethy.sdf`.  Most things work as expected. I had to move the MassShifter test deeper underwater as the vehicle would not pitch enough at the surface (kind of expected I guess?). I moved the mass shifter test deeper underwater.


chapulina (approved):
Working well for me!

PR #163, opened 2022-03-02
State message: populate UVW and PQR rates

chapulina:
Wrapping up #48 :checkered_flag: :raised_hands: :smiling_face_with_tear: 

This PR populates `ratePQR` and `rateUVW`, which are the quantities expressed in the FSK frame.

Note that the model itself is oriented as "SFU" (starboard, forward, up). See discussion on #48 on why it was tough to orient the model as FSK and save the extra transform. One way to visualize the SFU frame is using the transform control tool:

**ENU** :point_down: .................................................................. **FSK** :point_down:  / **SFU** :sparkles: .....................................................................  **NED** :point_down: 
sfu


arjo129 (approved):
Thanks! This looks good to me.

PR #171, opened 2022-03-15
update test fixture to report faults instead of errors

braanan:
@chapulina, @arjo129 Hi Luise and Arjo,

This is a minor change to the test fixture to output more relevant error info from the lrauv-application during testing. In the lrauv-app only FAULT and CRITICAL logging priorities are important.

Thanks!


arjo129 (approved):
Looks good to me.

PR #175, opened 2022-03-23
Use latest Docker image instead of garden

chapulina:
The garden image is outdated


braanan (approved):

chapulina:
I'll merge this to use the available image, but note that the `test_mission_pitch_and_depth_mass_vbs` test is failing

Issue #132, opened 2022-01-05
Install server configs instead of repeating plugins for each world


chapulina:
We have multiple worlds in this repository, each listing its own server `<plugin>`s. Instead of all the duplication, we could install one (or more) `.config` files with the set of plugins, and reuse that for every world. See this tutorial:

https://ignitionrobotics.org/api/gazebo/6.3/server_config.html

PR #182, opened 2022-03-28
Hoist ign-sensors IMU and magnetometer plugins to world SDFs

hidmic:
System plugins for IMU and magnetometer sensors are scene-wide plugins.


chapulina (approved):
Ahh we really should put some time into 

* https://github.com/osrf/lrauv/issues/132

braanan (approved):

braanan:
Hoping we can slowly get to a point where we only need to use empty_env...

hidmic:
Failing integration tests are unrelated to the AHRS. Merging!

PR #183, opened 2022-03-29
Separates the range bearing messages into two separate entities. 

arjo129:
Previously we were using the same RangeBearing message for both external communication and internal communication. This leads to lots of confusion as we were using the same message in two different places. This PR also changes the public facing `LRAUVRangeBearingResponse` to stop using a vector to store the coordinates, but rather explicitly label the coordinates.

### DIFF
--- a/lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
+++ b/lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
@@ -33,8 +33,8 @@ message LRAUVRangeBearingResponse
   /// \brief Request ID. Each request ID must be unique.
   uint32 req_id = 1;
 
-  /// \brief Bearing in terms of (r, theta, phi). This is based on ground truth
-  /// and not calculated using time-of-flight
+  /// \brief Bearing in terms of (r, elevation, azimuth). This is based
+  /// on ground truth and not calculated using time-of-flight
   ignition.msgs.Vector3d bearing = 2;
 
   /// \brief Time-Of-Flight reading. This is good for studying errors arising
--- a/lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
+++ b/lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
@@ -33,11 +33,16 @@ message LRAUVRangeBearingResponse
   /// \brief Request ID. Each request ID must be unique.
   uint32 req_id = 1;
 
-  /// \brief Bearing in terms of (r, elevation, azimuth). This is based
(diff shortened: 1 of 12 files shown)


lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
```diff
@@ -33,11 +33,16 @@ message LRAUVRangeBearingResponse
   /// \brief Request ID. Each request ID must be unique.
   uint32 req_id = 1;
 
-  /// \brief Bearing in terms of (r, elevation, azimuth). This is based
-  /// on ground truth and not calculated using time-of-flight
-  ignition.msgs.Vector3d bearing = 2;
+  /// \brief Range in terms of ground truth meters.
+  double ground_truth_range = 2;
+
+  /// \brief elevation in radians
```
braanan, line 39: `/// \brief vehicle referenced elevation in radians`

lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
```diff
@@ -33,11 +33,16 @@ message LRAUVRangeBearingResponse
...
+  double elevation = 3;
+
+  /// \brief azimuth in radians
```
braanan, line 42: `/// \brief vehicle referenced azimuth in radians`

lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
```diff
@@ -33,11 +33,16 @@ message LRAUVRangeBearingResponse
...
+  double azimuth = 4;
 
   /// \brief Time-Of-Flight reading. This is good for studying errors arising
   /// due to movement. But at high dt this may be prone to error.
-  double range = 3;
+  double range = 5;
```
braanan, line 47: @arjo129 replace with actual time-of-flight in seconds. We could also consider publishing the speed of sound through water that was used for this solution (fixed now, but could be calculated based on water density in the future)

lrauv_ignition_plugins/proto/lrauv_range_bearing_response.proto
```diff
@@ -33,11 +33,16 @@ message LRAUVRangeBearingResponse
   /// \brief Request ID. Each request ID must be unique.
   uint32 req_id = 1;
```
braanan, line 34: I think this message should also contain the local and remote addresses

braanan (changes requested):

arjo129:
Closing as acoustic comms is now working.

PR #174, opened 2022-03-23
Fixes acoustic range bearing orientation.

arjo129:
This PR fixes the range bearing transponder to follow the conventions sent to us by MBARI. In particular the range bearing system should calculate range-bearing in the following manner:
image

This PR introduces a new link that orients the acoustic transponder in the correct direction. Additionally it switches from traditional spherical coordinates to the one in the above diagram. 

We also have added a new test and example world with updated transponders. This test tests the correctness of the range-bearing program with respect to the model. The transponder placements are shown here:
transponder_labels

PR #185, opened 2022-03-31
Fix test_range_bearing.cc

hidmic:
Follow-up after #174. Alternatively, we could drop this test, as it is redundant with `test_acoustic_comms_orientation.cc`.


arjo129:
Thanks for spotting this. I think dropping this would be the way to go forward.

arjo129 (approved):

PR #187, opened 2022-04-05
Make integration tests pass

hidmic:
Precisely what the title says. This patch, coupled with some configuration changes for the LRAUV executable to avoid a go-to-surface behavior at the end of each test, makes integration tests pass.

### DIFF
--- a/lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
+++ b/lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
@@ -46,9 +46,6 @@ TEST_F(LrauvTestFixture, DepthVBS)
   EXPECT_NEAR(0.0, this->tethysPoses.back().Rot().Pitch(), 1e-6);
   EXPECT_NEAR(0.0, this->tethysPoses.back().Rot().Yaw(), 1e-6);
 
-  // Run non blocking
-  this->fixture->Server()->Run(false, 0, false);
-
   // Launch mission
   std::atomic<bool> lrauvRunning{true};
   std::thread lrauvThread([&]()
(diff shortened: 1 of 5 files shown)


hidmic:
FWIW I'm not convinced we should do this. I think we should fix hydrodynamics and tune control loops until the vehicle behaves in a physically sound way, but I'm not going to get in the way if we think it's best to move forward with it as-is.

lrauv_ignition_plugins/test/test_mission_depth_vbs.cc
```diff
@@ -101,16 +86,19 @@ TEST_F(LrauvTestFixture, DepthVBS)
     }
   }
 
-  // Vehicle's final pose should be near the 10m mark
-  EXPECT_NEAR(tethysPoses.back().Pos().Z(), -10, 1e-1);
+  // FIXME(hidmic): hydrodynamic damping forces induce
```
arjo129, line 89: As I mentioned we should not fix this as it will require re-parameterization. I'm hoping adding the cross terms will make this go away.

lrauv_ignition_plugins/test/test_mission_pitch_and_depth_mass_vbs.cc
```diff
@@ -88,19 +88,19 @@ TEST_F(LrauvTestFixture, PitchDepthVBS)
   for (const auto pose: this->tethysPoses)
   {
     // Vehicle should dive down.
-    EXPECT_LT(pose.Pos().Z(), 0.1);
+    EXPECT_LE(pose.Pos().Z(), 0.5);
```
arjo129, line 91: This makes sense as there is minor oscillation.

lrauv_ignition_plugins/test/test_mission_pitch_mass.cc
```diff
@@ -89,19 +89,27 @@ TEST_F(LrauvTestFixture, PitchMass)
   // Euler.Z = yaw
   for(auto pose: this->tethysPoses)
   {
-    // Max Pitch 20 degrees. Allow 5 degrees error for oscillations.
-    // Min Pitch -5 degrees. Again allow 5 degrees error for oscillation.
+    // FIXME(hidmic): when surfacing, the vehicle oscillates
```
arjo129, line 92: I think the problem here is we are testing at the surface which makes behaviors hard to define and will into oscillations. Rather than tweaking this parameter, we should not derive from `LrauvTestFixture` but have the test derive from [`LrauvTestFixtureAtDepth`](https://github.com/osrf/lrauv/blob/b84780f0cb3848b84f0e454fd0fdf2dc2b30241e/lrauv_ignition_plugins/test/helper/LrauvTestFixture.hh#L320). I thought I had done this but its not clearly the case.

arjo129 (changes requested):
I think for PitchMass we should not test at the surface which is what `LrauvTestFixture` does we should instead use `LrauvTestFixtureAtDepth`

lrauv_ignition_plugins/test/test_mission_pitch_and_depth_mass_vbs.cc
hidmic: It's not the oscillation, but the fluid density changing at z = 0.5m. I was meaning to ask, should the interface be at -0.5m instead?

lrauv_ignition_plugins/test/test_mission_pitch_mass.cc
hidmic: On a second inspection, we still get an initial oscillation about pitch = 0 if running the test at a depth (which requires it to be neutrally buoyant) as soon as the LRAUV controller is engaged. I'll check for spurious mass shifter commands.

arjo129:
Hmm I wonder if the default behviour ben set is commanding it to go to the
surface first.

hidmic:
> Hmm I wonder if the default behviour ben set is commanding it to go to the
surface first.

It is. We can disable it for tests though (I dropped it from the default behavior, we can drop it from the startup behavior too).

arjo129:
That might cause a bit of disturbance at the start for pitch-mass I think.

hidmic:
Superceded by #189, #192 and its offspring.

PR #188, opened 2022-04-06
Add a VolumetricScalarField with support for interpolation. 

arjo129:
This PR introduces a faster lookup method that is very similar to the one used by the original LRAUV simulator. It assumes that the data is in some form of a grid. This object is called the `VolumetricScalarField`. This allows for extremely fast data lookup as oppsed to a KNN search. 

## TODO
- [ ] write tests


arjo129:
Superseded by #190

Issue #189, opened 2022-04-06
Revisit `lrauv_ignition_plugins` testing suite


hidmic:
## Motivation

Integration tests have been broken for a while now. Recent attempts to fix them have been met with modelling limitations, upstream changes, inadequate tuning, etc (see #187). To optimize our time, we need to clearly define what use cases we want to validate in CI, update our tests to match expectations, and ensure our code base passes.

## Definition of Done

- [x] Review existing `lrauv_ignition_plugins` tests
- [x] Define/discuss tests to match use cases
- [ ] Update tests to match definitions
- [ ] Make tests (and CI) pass

hidmic:
FYI @arjo129. I take that YoYoCircle, PitchMass at a depth, and PitchAndDepthMassVBS are a must (at the bare minimum).

arjo129:
Hey @hidmic  this set of requirements is very well documented. I think you're spot on with the Mission Control tests I'll just throw my two cents of why every test was written the way it was just to check if we are not missing anything.

## lrauv_description
The tests here verify that the vehicle model is hydrostatically stable. The vehicle has a slight offset between the center of buoyancy and center of mass to ensure a restoring moment.

## lrauv_ignition_plugins
These tests guarantee hydrostatic stability.
* `TEST(Stability, NeutralBuoyancy)` - Checks that the vehicle is actually neutrally buoyant. We need to be very careful here because small floating point error propagates.
* `TEST(Stability, RestoringMoment)` - Makes sure that there is a restoring moment.We start the vehicle at a pitch and the vehicle should behave like a simple oscillator as the restoring moment will kick in there is no hydrodynamics enabled to damp oscillations.

### Vehicle dynamics tests
This set of tests checks the behavior of our vehicle's physics in ignition.
We don't integrate with the controller here as these help us identify bugs
and short comings of our physics system.
* `test_buoyancy_action.cc` - Unit test that checks the buoyancy engine can inflate and deflate and the vehicle heads in the correct direction (up/down). Known issue: Significant lateral translation due to parameterization of Fossen's equations.
* `test_controller.cc` - Check tethyscomms plugin (in charge of bridging ignition to controller) actually loads.
* `test_drop_weight.cc` - Tests that the drop weight actually works and causes the
vehicle to float up.
* `test_elevator.cc`- Check the elevator actually controls the direction of motion. Also checks the sign convention of the actuators. Elevators use the LiftDrag plugin.
[...]

PR #192, opened 2022-04-20
Rework `lrauv` test suite

hidmic:
Precisely what the title says. Following #189 and building on top of https://github.com/osrf/lrauv/pull/187, this patch aggregates all tests in an `lrauv_system_tests` package downstream of `lrauv_ignition_plugins` and `lrauv_description`. This package also includes useful abstractions, derived from existing test patterns, with which all tests have been reworked to reduce duplication and flakyness. 

All tests but mission tests have been verified to pass locally. Missions tests will be checked next (as custom MBARI configuration is required). A README.md for the `lrauv_system_tests` package summarizing #189 is yet to be added too.


hidmic:
Note for the reviewers: each commit affects a single package and thus it would be straightforward to split this PR if necessary.

hidmic:
Alright, all tests are passing locally on this patch. I'll go ahead and split it while new MBARI images get built.

hidmic:
@braanan @arjo129 this patch contains #193 through #199 combined. Might be useful during review.

PR #193, opened 2022-04-21
Rework `lrauv_description` package

hidmic:
This PR reworks the `lrauv_description` package in preparation for the `lrauv_system_tests` package. Namely, it:

- Exports CMake targets for downstream usage
- Adds colcon hook equivalent for sh-like shells
- Removes all tests (migrated to `lrauv_system_tests`)


braanan (approved):
This PR requires fixes in #194 to build so I suggest we merge them together.

PR #194, opened 2022-04-21
Rework `lrauv_ignition_plugins` package

hidmic:
This PR reworks the `lrauv_ignition_plugins` package in preparation for the `lrauv_system_tests` package. Namely, it:

- Exports CMake targets for downstream usage
- Adds colcon hook equivalent for sh-like shells
- Removes all tests (migrated to `lrauv_system_tests`)


arjo129 (changes requested):
Ideally this PR should have tests passing since there are no unit tests. I think you are missing the python bindings dependency. See #200.

arjo129 (approved):

hidmic:
Thank you for the review and the quick fix @arjo129 !

braanan:
@hidmic Looks like this PR will need one more adjustment to catch up with the recent ign version updates (cmake2 -> cmake3) — should I push this fix here or in some other PR?

braanan:
Went for it.

braanan (approved):

PR #195, opened 2022-04-21
Add lrauv_system_tests package.

hidmic:
Including support libraries for testing. Depends on #193 and #194.

### DIFF
--- a/lrauv_system_tests/include/lrauv_system_tests/Client.hh
+++ b/lrauv_system_tests/include/lrauv_system_tests/Client.hh
@@ -1,0 +1,106 @@
+/*
+ * Copyright (C) 2022 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+/*
+ * Development of this module has been funded by the Monterey Bay Aquarium
+ * Research Institute (MBARI) and the David and Lucile Packard Foundation
+ */
+
+#ifndef LRAUV_SYSTEM_TESTS__CLIENT_HH
+#define LRAUV_SYSTEM_TESTS__CLIENT_HH
+
+#include <future>
+#include <mutex>
+#include <string>
+#include <unordered_map>
+
+#include <ignition/common/Console.hh>
+#include <ignition/transport/Node.hh>
+
+namespace lrauv_system_tests
+{
+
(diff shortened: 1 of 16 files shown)


lrauv_system_tests/include/lrauv_system_tests/Subscription.hh
```diff
@@ -0,0 +1,137 @@
...
+  /// message history buffer. Buffer grows indefinitely
+  /// by default.
+  public: void Subscribe(
+      ignition::transport::Node &_node,
+      const std::string &_topicName,
+      size_t _messageHistoryDepth = 0u)
+  {
+    if (this->subscribed)
+    {
+      throw std::logic_error("cannot ");
```
arjo129, line 55: I think the error message got cut short. Also might be worth making this the constructor rather than having a separate `Subscribe()`.

arjo129 (approved):
Thanks for this awesome refactor. I have a few minor knits here and there but other than that this refactor looks much better than whatever we had earlier. I can see many of these classes being very useful.

lrauv_system_tests/include/lrauv_system_tests/Client.hh
```diff
@@ -0,0 +1,106 @@
...
+      throw std::runtime_error("Failed to subscribe to " + responseTopicName);
+    }
+    this->requestsPublisher =
+        _node.Advertise<RequestMessageT>(_ns + "/requests");
+  }
+
+  /// Request a response from the remote server.
+  /// \param[in] _request Request message instance.
+  /// \returns a future response message, to be awaited.
+  public: std::future<ResponseMessageT> Request(RequestMessageT _request)
```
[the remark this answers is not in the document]
hidmic: Not `const`, I have to mutate it right below on line 64. It _could_ be a mutable reference though. I have a slight preference towards not mutating the caller's scope (it's less surprising that way), but I can be convinced otherwise.

lrauv_system_tests/include/lrauv_system_tests/Subscription.hh
hidmic: > I think the error message got cut short.

Yeap! Good catch. Fixed in https://github.com/osrf/lrauv/pull/195/commits/34cfeeba428b0b0e175976fc0dba48f8df6753a6.

>  Also might be worth making this the constructor rather than having a separate `Subscribe()`.

About that. I agree with you, I don't like the half constructed object. However, depending on how you write tests, having a default constructor can be useful. E.g. [here](https://github.com/osrf/lrauv/pull/192/files#diff-679ef17c89eb7e8b08f01621e4846008baafc06f4100bd53e70d4af5cbf0a53bR288-R289), I'd have to cram everything in a field initializer or dynamically allocate the object to defer construction if there wasn't a `Subscribe()`.

lrauv_system_tests/include/lrauv_system_tests/Client.hh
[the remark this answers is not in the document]
arjo129: That's fine. In any case this is stuff that we can change dynamically.

lrauv_system_tests/include/lrauv_system_tests/Subscription.hh
arjo129: Fair enough. Makes sense.

hidmic:
Alright, all green and reviewer's happy. Merging.

PR #200, opened 2022-04-25
Bump ign-utils and add dependency on ignition python bindings

arjo129:
I noticed that our CI has been failing because ign-utils has undergone a version bump and our container did not have the appropriate ignition python bindings. This PR bumps the relevant versions and adds the dependency on ignition's python bindings.


docker/tests/Dockerfile
```diff
@@ -48,7 +48,7 @@ RUN /bin/sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable
 
 # Install and upgrade the latest Ignition binaries
 RUN apt-get -qq update && apt-get -q -y install \
-  ignition-garden && apt-get dist-upgrade -y
+  ignition-garden python3-ignition-gazebo7 python3-ignition-math6 && apt-get dist-upgrade -y
```
hidmic, line 51: @arjo129 fix:
```suggestion
  ignition-garden python3-ignition-gazebo7 python3-ignition-math7 && apt-get dist-upgrade -y
```
Garden brings Ignition Math 7.

lrauv_ignition_plugins/CMakeLists.txt
```diff
@@ -38,8 +38,8 @@ set(IGN_MSGS_VER ${ignition-msgs9_VERSION_MAJOR})
 find_package(ignition-plugin1 REQUIRED COMPONENTS register)
 set(IGN_PLUGIN_VER ${ignition-plugin1_VERSION_MAJOR})
```
hidmic, line 39: @arjo129 I believe it should be:
```suggestion
find_package(ignition-plugin2 REQUIRED COMPONENTS register)
set(IGN_PLUGIN_VER ${ignition-plugin2_VERSION_MAJOR})
```
instead. Pushed the same fix in https://github.com/osrf/lrauv/pull/194.

hidmic (commented):
Thanks for the fix. There's overlap with https://github.com/osrf/lrauv/pull/194 though. Don't mind which ones goes in first though.

PR #206, opened 2022-05-09
Fix Portuguese Ledge Example

arjo129:
I was trying to run the example and ran into two problems. First was that the SDFormat file we generated was invalid. Second, I found the file was not being installed to the `worlds` directory, hence we were unable to run `ign gazebo portuguese_ledge.sdf`. This PR fixes these problems.


lrauv_ignition_plugins/CMakeLists.txt
```diff
@@ -269,4 +273,4 @@ install(
     models
     worlds
     ${CMAKE_CURRENT_BINARY_DIR}/hooks
-  DESTINATION share/${PROJECT_NAME})
+  DESTINATION share/${PROJECT_NAME})
```
hidmic, line 276: @arjo129 nit: unrelated change.

hidmic (approved):
Must test locally, but changes LGTM.

PR #208, opened 2022-05-11
Adds a simple panel for loading custom science data 

arjo129:
This is a minor quality of life improvement but it enables science data to be loaded using a GUI. You can click on "Environmental Configuration" panel when running `ign gazebo empty_environment.sdf` then it will open a file browser to load the correct data. Under the hood this will sens a message the to science data plugin to trigger a reload.
image


arjo129:
Closed in favor of #211

PR #209, opened 2022-05-11
Adds a plugin which spawns tethys

arjo129:
This PR adds a plugin which spawns a copy of the tethys vehicle at a given location.

spawn_lrauv


hidmic (approved):
Works as intended, though the plugin picking up the first spawned vehicle as the world origin is slightly misleading at first. Maybe document the behavior? Use a tooltip?

Issue #159, opened 2022-02-16
Investigate why test fixtures go :boom: if we have more than one TEST in a single process.


arjo129:
Google protobuf doesn't like the way we are linking the messages in our repository. Whenever we try to run two TestFixtures in the same process the tests will segfault.

hidmic:
https://github.com/protocolbuffers/protobuf/issues/4126 suggests loading/unloading protobuf message shared libraries leaves bad global state behind. Perhaps we can preload them?

arjo129:
Hopefully eventually we can move away from custom messages then?

marcoag:
Dunno if related but I also had troubles trying to run two servers in the same test in `gz-sim` so I have to split them in two files ([here](https://github.com/gazebosim/gz-sim/blob/ign-gazebo6/test/integration/model_photo_shoot_default_joints.cc) and [here](https://github.com/gazebosim/gz-sim/blob/ign-gazebo6/test/integration/model_photo_shoot_random_joints.cc)).

hidmic:
We know why this happens and it's very much core to how `protobuf` works. I'm inclined to close this as `wont fix`.

PR #210, opened 2022-05-13
Adds time interpolation to the Science Sensors

arjo129:
As said above. There are several major changes in this PR.
* Move all the logic for lookup in the science sensor to a separate function `ScienceSensorsSystemPrivate::InterpolateInTime()`.
* Add time interpolation.
* Add the ability to load files via a topic instead of hard-coded into the SDFormat world.
* Add unit tests for time interpolation.


lrauv_system_tests/test/vehicle/test_sensor_timeinterpolation.cc
```diff
@@ -0,0 +1,171 @@
+/*
```
hidmic, line 1: @arjo129 meta: why a separate source file, as opposed to adding a test case to `test_sensor.cc`?

lrauv_system_tests/test/vehicle/test_sensor_timeinterpolation.cc
```diff
@@ -0,0 +1,171 @@
...
+  {
+    std::this_thread::sleep_for(100ms);
+    // Run paused so we avoid the physics moving the vehicles
+    fixture->Server()->RunOnce(true);
+    expectedIterations++;
+    igndbg << "Waiting for vehicles to spawn" << std::endl;
+  }
+  EXPECT_TRUE(spawnedAllVehicles);
+
+  fixture->Server()->Run(true, 1000, false);
```
hidmic, line 170: @arjo129 meta: there are a number of tools in `lrauv_system_tests` to reduce duplication, prevent races, reduce total run time, and improve readability across tests. I'd be inclined to use them if time allows. [`test_spawn.cc`](https://github.com/osrf/lrauv/blob/main/lrauv_system_tests/test/simulation/test_spawn.cc) is one good example.

lrauv_system_tests/test/vehicle/test_sensor_timeinterpolation.cc
arjo129: https://github.com/osrf/lrauv/issues/159 I cant load multiple test cases with the auv without :boom: happening.

hidmic: Ahh, that's right.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -634,22 +744,23 @@ void ScienceSensorsSystem::PostUpdate(const ignition::gazebo::UpdateInfo &_info,
       return;
     }
   }
+  else{
+    std::lock_guard<std::mutex> lock(this->dataPtr->dataMutex);
+    if (this->dataPtr->newDataAvailable)
```
hidmic, line 749: @arjo129 consider making `newDataAvailable` an `std::atomic_bool` to avoid locking a mutex on every simulation step. This flag will evaluate to `false` most of the time.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
```diff
@@ -52,9 +53,28 @@ class tethys::ScienceSensorsSystemPrivate
...
+    igndbg << "Reloading file " << _filepath.data() << "\n";
+
+    // Trigger reload and reread data
+    std::lock_guard<std::mutex> lock(this->dataMutex);
+    this->newDataAvailable = true;
+    this->dataPath = _filepath.data();
+  }
+
+  /// \brief mutex for updating data.
+  public: std::mutex dataMutex;
```
hidmic, line 69: @arjo129 hmm, isn't this mutex redundant with `mtx`?

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
arjo129: To the best of my knowledge that would lead to incorrect behavior. Suppose multiple requests come in on the subscription topic, if this were purely an atomic there is no guarantee that the filepath will not have changed mid way through our read process. We may end up with a scenario where the latest file is not loaded.

lrauv_ignition_plugins/src/ScienceSensorsSystem.cc
arjo129: No mtx only protects reading. This protects the filename as well as whether data is ready. We could elminate `mtx` since it does not really server any purpose at this point.

hidmic (approved):
LGTM -- pending comments are not blockers.

PR #212, opened 2022-05-13
Add DVLVelocityTracking message

hidmic:
Including nested messages. I haven't even tried to build this yet but I wanted to put it out for you guys to see as fast as possible. I think this message hierarchy should provide more than enough flexibility while staying semantically meaningful.


lrauv_ignition_plugins/proto/dev_beam_state.proto
```diff
@@ -0,0 +1,45 @@
...
+  DVLKinematicEstimate velocity = 2;
+
+  /// \brief Beam range estimate
+  DVLRangeEstimate range;
+
+  /// \brief Beam signal strength indicator
+  optional double rssi = 6;
+
+  /// \brief Measured background noise spectral density
+  optional double nsd = 7;
```
hidmic, line 41: Maybe this could be aggregated in a message about quality metrics?

lrauv_ignition_plugins/proto/dvl_velocity_tracking.proto
```diff
@@ -0,0 +1,59 @@
...
+  enum DVLType
+  {
+    PISTON = 0;
+    PHASED_ARRAY = 1;
+  };
+  /// \brief Type of DVL
+  DVLType type = 2;
+
+  /// \brief Locked on target
+  DVLTrackingTarget target = 3;
```
hidmic, line 49: Hmm, here you can only specify one tracking target (i.e. tracking mode). IIUC DVLs can do both simultaneously. We could cope with that using different topics, but we can also refactor this to include both (surely not having to synchronize them downstream would be best).

lrauv_ignition_plugins/proto/dev_beam_state.proto
```diff
@@ -0,0 +1,45 @@
+/*
```
arjo129, line 1: Knit: File name is `dev_beam_state.proto` would probably be better to call it `dvl_beam_state.proto`

lrauv_ignition_plugins/proto/dvl_kinematic_estimate.proto
```diff
@@ -0,0 +1,41 @@
...
+     DVL_REFERENCE_BEAM = 2;
+  };
+  /// \brief Estimate frame of reference (incl. conventions)
+  ReferenceType reference = 1;
+
+  /// \brief Estimate mean
+  Vector3d mean = 2;
+
+  /// \brief Estimate covariance
+  repeated double covariance = 3;
```
arjo129, line 40: Knit: document the expected covariance matrix fromat error.

lrauv_ignition_plugins/proto/dvl_velocity_tracking.proto
arjo129: Will we be estimating this at all?

arjo129 (commented):
Just some minor comments

lrauv_ignition_plugins/proto/dvl_velocity_tracking.proto
hidmic: Well, there will always be a target: either a water mass or the sea floor. The amount of data we'll provide may vary (and thus why several fields are optional).

hidmic:
Looks like the test failure in https://github.com/osrf/lrauv/actions/runs/2463373105/attempts/2 was a flake. I cannot reproduce it locally at all, and CI's passing now. 

I suspect the thresholds we have in mission tests to cope with hydrodynamic modelling issues. Those thresholds were empirically set, so it may be that we exceed them from time to time. Ultimately, we have to fix the underlying issues for this go away completely.

PR #214, opened 2022-05-31
Migrate from `ignition` to `gz`

hidmic:
Precisely what the title says. Ignition is now Gazebosim, and we need to accomodate changes.

### DIFF
--- a/lrauv_ignition_plugins/src/TethysCommPlugin.cc
+++ b/lrauv_ignition_plugins/src/TethysCommPlugin.cc
@@ -22,20 +22,20 @@
 
 #include <chrono>
 
-#include <ignition/gazebo/Util.hh>
-#include <ignition/gazebo/components/AngularVelocity.hh>
-#include <ignition/gazebo/components/JointPosition.hh>
-#include <ignition/gazebo/components/JointVelocity.hh>
-#include <ignition/gazebo/components/LinearVelocity.hh>
-#include <ignition/gazebo/components/Pose.hh>
-#include <ignition/msgs/double.pb.h>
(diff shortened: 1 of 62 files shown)


hidmic:
This only migrates namespaces in C++ API and Protobuf messages, enough to get it building with latest Garden. Shall we move forward with the rest now? Ie. replace the `ignition` term for `gz` everywhere.

hidmic:
CI will fail until we've migrated the `LRAUV` controller codebase (and update our Docker image).

lrauv_ignition_plugins/src/TethysCommPlugin.cc
```diff
@@ -260,59 +260,59 @@ void TethysCommPlugin::Configure(
...
 
-  this->elevatorTopic = ignition::transport::TopicUtils::AsValidTopic(
+  this->elevatorTopic = gz::transport::TopicUtils::AsValidTopic(
     "/model/" + _ns + "/joint/" + this->elevatorTopic);
   this->elevatorPub =
-    this->node.Advertise<ignition::msgs::Double>(this->elevatorTopic);
+    this->node.Advertise<gz::msgs::Double>(this->elevatorTopic);
   if (!this->elevatorPub)
   {
     ignerr << "Error advertising topic [" << this->elevatorTopic << "]"
```
chapulina, line 289: Btw, you should be able to migrate these console macros already if you want to. They aren't giving warnings right now though.

chapulina (approved):
> replace the ignition term for gz everywhere.

Note that migration is still in progress upstream and you may not be able to migrate everything just yet

chapulina (approved):

PR #230, opened 2022-07-18
Relocate DVL to payload section

hidmic:
Precisely what the title says. Follow-up after #213.


arjo129 (approved):
LGTM

hidmic:
@arjo129 PTAL, there were a few _details_ lurking below the surface.

hidmic:
Alright, our frame conventions are a bit of a mess.I'm going for minimal changes. We can rework them in the future.

hidmic:
Yay, finally. Going in, in the interest of time.

PR #233, opened 2022-08-10
Adds a flag to enable/disable mission tests

arjo129:
The idea is to enable rapid iteration on the opensource front without relying on the mbari controllers.


lrauv_system_tests/CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ find_package(gz-math7 REQUIRED)
 find_package(gz-transport12 REQUIRED)
 find_package(lrauv_ignition_plugins REQUIRED)
 
+# Add option to build without LRAUV integration
+option(ENABLE_MISSION_TESTS
+  "Enable mission tests. (Requires integration with MBARI's controllers)"
+  ON)
```
hidmic, line 18: @arjo129 I think it makes sense to disable this by default, as the vast majority of potential downstream users won't have access to MBARI code.

hidmic (approved):

PR #239, opened 2022-08-16
Use custom gz-garden base image for CI

hidmic:
Precisely what the title says. Went for manual dispatch to keep it simple.


adityapande-1995 (approved):
Hmm, CI says access is denied ?

adityapande-1995 (commented):
Also, as per our discussion earlier, should we change the branch names in repos file to "lrauv-garden" or something like that ?

adityapande-1995 (approved):

PR #250, opened 2022-09-21
Use upstream sensor infrastructure [WIP]

arjo129:
This PR is trying to migrate our code to use the upstream sensors code.


arjo129:
Test failures seem unrelated.

hidmic (commented):
Looks reasonable to me, once we've sorted out https://github.com/gazebosim/gz-sim/pull/1814

arjo129:
There are issues with using the ECM for loading stuff via GUI. The current system works and given the limited time we have, I think we should not risk breaking things.

PR #252, opened 2022-09-23
Removes `.kate-swp` file

arjo129:
Just noticed a kate-swp file in the repo. I don't think its meant to be there. Adding to our .gitignore.


hidmic (approved):
LGTM

PR #255, opened 2022-09-30
Increase tolerance in AHRS tests

adityapande-1995:
This should fix the flaky failures in AHRS tests.

Signed-off-by: Aditya <aditya050995@gmail.com>


hidmic (approved):
Tight. Should do.

PR #258, opened 2022-10-07
Switching to gazebo upstream hydrynamics plugin

marcoag:
Depends on https://github.com/gazebosim/gz-sim/pull/1692 getting ported to `gz-sim7`.

This PR adds the changes so we can start using the upstream hydrodynamics plugin and removes the local one from this repo.


arjo129:
Lots of failing tests here. I will take a look at this.

arjo129:
Once https://github.com/gazebosim/gz-sim/pull/1794 is merged and we switch back to using the main branch, we should be good to merge this.

arjo129 (approved):
Failing tests are:
```
         9 - VehicleStateTest.ThrustState (Failed)
	 10 - VehicleStateTest.ThrustAndTurnState (Failed)
	 16 - AHRSTest.FrameConventionsAreCorrect (Failed)
	 24 - BuoyancyActionTest.Sink (Failed)
	 27 - MassShifterTest.DownwardTilt (Failed)
```
* 9: the error is here: https://github.com/osrf/lrauv/actions/runs/3694474851/jobs/6255926270#step:4:778. Seems like an issue of tolerance
* 10: the error is here: https://github.com/osrf/lrauv/actions/runs/3694474851/jobs/6255926270#step:4:882. Also seems like a tolerance issue.
* ~~16. Just seems flaky :skull_and_crossbones:  (It fails on other PRs as well). Unrelated to dynamics.~~
* 24: Seems like decreasing buoyancy causes the vehicle to not only sink, but also spin. TBH this is a known issue with the set of params for the LRAUV. It seems to have gotten worse after adding cross term support (likely due to numerical stability issues). To be honest theres nothing to worry about here because the controller should correct for such unwanted behaviour.
* 27: Also seeing some spinning here like in 24. Similar to 27.

Overall, we see slight regression in the numerical correctness of the equations. On the other hand, we should still be able to control the vehicle. Hence, I'm approving this. If we want we can widen tolerances on the aforementioned tests. Also the hydrodynamics tests so I'm happy.

hidmic:
> 	 16 - AHRSTest.FrameConventionsAreCorrect (Failed)

Should be fixed by #276.

arjo129:
Closing because we don't have enough time. Its safer to guarantee that things work. Since the hydrodynamics here works.

PR #260, opened 2022-10-13
Adds a friendly warning if the python bindings for gazebo are not found

arjo129:
While installing gazebo on a new computer, I ran into trouble with python gazebo bindings.
I believe part of the issue is I have gazebo newly compiled from source. It's not great that
the repo does not compile at all if gz-python is not found, especially since gz-python is not
used by any core dependency. This PR adds a simple check and only runs the world generation
if the python-gazebo package is found. If the package is not found, users can still install the
simulator, however a warning is given warning users that the examples may not work.

I'm only checking for gz, but it might be worth checking for other python packages which may
or maynot come installed with an end user's work space.

### DIFF
--- a/lrauv_gazebo_plugins/CMakeLists.txt
+++ b/lrauv_gazebo_plugins/CMakeLists.txt
@@ -51,6 +51,11 @@ find_package (Eigen3 3.3 REQUIRED)
 
 find_package(PCL 1.2 REQUIRED)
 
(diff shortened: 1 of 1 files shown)


hidmic (approved):
LGTM, though I wonder if we shouldn't wrap this template expansion in a macro for improved readability.

PR #268, opened 2022-11-03
Migrates the hook files

arjo129:
We are still using the old IGN environment variables. This PR migrates them to gz.


hidmic (approved):
Tested locally

PR #269, opened 2022-11-03
Removes `<namespace>` in joints.

arjo129:
I have no idea why there are namespaces in joints. To the best of my knowledge this is not part of the SDFormat spec and it seems that gazebo is complaining about it.


hidmic:
@adityapande-1995 here we have an instance of a multi-vehicle test fluke https://github.com/osrf/lrauv/actions/runs/3383343331/jobs/5619146123.

hidmic (approved):
Tested locally.

PR #272, opened 2022-11-14
Clean up `lrauv_gazebo_plugins` worlds

hidmic:
Precisely what the title says. Incidentally closes https://github.com/osrf/lrauv/issues/132.


hidmic:
@arjo129 @adityapande-1995 @caguero I'm seeing strange behavior. Running:

```sh
gz sim -v4 --gui-config install/share/lrauv_gazebo_plugins/config/gui.config portuguese_ledge.sdf
```

has the intended outcome, whereas:

```sh
gz sim -v4 --gui-config install/share/lrauv_gazebo_plugins/config/gui.config tethys_at_portuguese_ledge.sdf
```

yields nothing. Do you see the same?

arjo129:
Neither are working for me. I'm getting the following error:
```
 [GUI] [Err] [Application.cc:290] Failed to load file [install/share/lrauv_gazebo_plugins/config/gui.config]: XMLError
[GUI] [Wrn] [Gui.cc:373] Failed to load config file[install/share/lrauv_gazebo_plugins/config/gui.config].
```

hidmic:
@arjo129 ahh, I wasn't explicit about how to build the workspace. It  should be merged installed !

arjo129:
Hmmm lots of weird behavior. I can see the gui but no hint of tiles or tethys.

arjo129:
This is what I see for both:
image
image

Might I add that I believe the hooks file assumes a non-merge-install scenario.

hidmic:
> Hmmm lots of weird behavior. I can see the gui but no hint of tiles or tethys.

It's less weird than having only one working :sweat_smile: I suspect some sort of race in how plugins are loaded, but it's just a hunch.

> Might I add that I believe the hooks file assumes a non-merge-install scenario.

Hmm, shouldn't be the case as we're using the right install prefix:
https://github.com/osrf/lrauv/blob/74389584ffd79cf62473c7f7be38c79d5da3ea32/lrauv_gazebo_plugins/hooks/hook.sh.in#L1-L4

hidmic:
Alright, it was picking up another SDF file with the same name for me... Now we are seeing the same.

hidmic:
Curiously enough, removing `ScienceSensorSystem` and `EnvironmentPreload` plugins "fixes" the problem. Maybe not a race but either plugin failing to setup?

arjo129:
We should be removing science sensor soon enough. I can take alook at EnvironmentPreload and see what happens.

hidmic:
I gave up on trying to figure out what is going. Merged everything back into each world in https://github.com/osrf/lrauv/pull/272/commits/7e72734532e1c736162406e7ac9ad1220f92a03b. That works just fine.

hidmic:
@caguero re: number of worlds. Let me know if you want to go away with the worlds that start with a Tethys in them or not. Remember there is no generic mechanism to work with these worlds (everything is setup to work with MBARI LRAUVs only).

arjo129:
Getting some weird behavior. Not sure exactly what went wrong. Need to investigate.
[Screencast from 11-22-2022 11:36:08 AM.webm](https://user-images.githubusercontent.com/542272/203215308-ba12dcfd-e739-4180-b3d6-43438a364914.webm)

arjo129:
Seems like if a vehicle is spawned, `EachNew()` is not catching it [here](https://github.com/gazebosim/gz-sim/blob/3d94c0ba2afbe0480a1dfccd9d9fdf94aa91b761/src/systems/buoyancy/Buoyancy.cc#L354). I think this seems to be an issue with the ECM.

hidmic:
> Seems like if a vehicle is spawned, EachNew() is not catching it [here](https://github.com/gazebosim/gz-sim/blob/3d94c0ba2afbe0480a1dfccd9d9fdf94aa91b761/src/systems/buoyancy/Buoyancy.cc#L354), hence buoyancy is not getting executed. I think this seems to be an issue with the ECM.

@arjo129 uh, good find. It is catching it though, but no collisions are found (at the time). A race?

arjo129:
OK so it makes sense. Seems like the `UserCommand` system responsible for creating the entities only creates the entities during the `PreUpdate` phase. At the same time the buoyancy plugin only handles new entities at creation time only during the `PreUpdate` phase. At the end of the time step new entities are all reset. Hence, if Buoyancy runs before `UserCommand` `EachNew()` will not catch anything new created and by the second time step all new creations will have been cleared. I have a fix for the buoyancy plugin that I will push now.

hidmic:
> Not the only unexpected thing that happened

Someone went to bed in a jokey mood.

caguero:
> @caguero re: number of worlds. Let me know if you want to go away with the worlds that start with a Tethys in them or not. Remember there is no generic mechanism to work with these worlds (everything is setup to work with MBARI LRAUVs only).

If we still don't have a generic way of launching vehicles I'm leaning towards keeping it as it is preserving the worlds.

arjo129 (approved):
Works for me once gazebosim/gz-sim#1808 is merged.

PR #275, opened 2022-12-13
Fix default docker command

hidmic:
Precisely what the title says. A follow-up after #272. An omission of mine.


caguero (approved):

hidmic:
The only test failure is unrelated to this patch. Merging, then looking into it.

PR #278, opened 2022-12-15
Adds cross terms

arjo129:
Do not merge.


arjo129:
Dont want to break the sim,

PR #282, opened 2023-06-30
Add paper citation to README

mabelzhang:
Add our paper citation so people can cross-reference easier.
DOI isn't out yet :woman_shrugging: Probably won't be for months.


arjo129 (approved):
Thanks!

